PR #132, opened 2016-12-06
Identify customer

seanrucker:
Adds the ability to identify a customer through the new traits API which is a set of instructions on how to modify the customer. This approach sets up well for the Rover 2.0 architecture.

Additionally the `Rover.identify` method now dispatches a device update event so that customer updates are persisted immediately. 

Similarly a `Rover.clearCustomer` method was added for "signing out".


chrisrecalis (approved):
:shipit:

PR #176, opened 2017-05-08
add support for remote urls that return a data uri image string

bboralRMN:
@seanrucker here's a stab at data uri support.
This is for our particular use-case where we pass Rover an image url like "barcode.com?id=5" and that returns a data uri string like this "data:image/png;base64,AAA"

### DIFF
--- a/Rover/UIImageView+RVAsset.swift
+++ b/Rover/UIImageView+RVAsset.swift
@@ -43,7 +43,19 @@ extension UIImageView {
             
             guard let data = data else { return }
             
-            self.image = UIImage(data: data)
+            if let img = UIImage(data: data) {
+                self.image = img
+            } else {
+                // data uri image support
+                do {
+                    if let str = String(data:data, encoding: String.Encoding.utf8),
+                        let dataUrl = URL(string: str),
+                        let decodedData = try? Data(contentsOf: dataUrl),
+                        let img = UIImage(data: decodedData) {
+                        self.image = img
+                    }
+                }
+            }
         }
     }
 }


seanrucker:
I thought #173 added everything you needed for barcodes? Are you generating the data URIs locally or you are calling a URL that returns another URL?

bboralRMN:
I didn't correctly describe the use-case to you, previously. We're not generating data uris locally. We're calling out to a url that returns the data uri.

seanrucker:
In that case can you call the URL and retrieve the data URI *before* you modify the JSON payload? Supporting data URIs in image URLs makes sense to me, but the modification you've submitted here is very specific to your use case. You'd be using the AssetManager for something it was not meant for.

seanrucker:
We are about to push the corresponding Android update to support data URIs tomorrow.

Rover/UIImageView+RVAsset.swift
```diff
@@ -43,7 +43,19 @@ extension UIImageView {
             
             guard let data = data else { return }
             
-            self.image = UIImage(data: data)
+            if let img = UIImage(data: data) {
+                self.image = img
+            } else {
```
seanrucker, line 48: This assumes that if it could not create an image from the data that the response payload must be a data URI. This is very specific to RMN's use case. 

The feature we added in #173 was to support data URIs in the JSON payload . We check the URL scheme and if it is indeed a valid data URI then we bypass the Asset manager (don't need to download anything) and render the image directly. That use case is much more general purpose and potentially useful to other users. I'm not sure we can accept this PR as it is.

PR #350, opened 2018-11-13
Feature/333 remove present view action

orospakr:
Depends on #349.

PR #351, opened 2018-11-13
Feature/332 remove open notification action

### LINKED ISSUE #332: Remove OpenNotificationAction
To support #329 we need to remove this action.

## Todo

- [x] Inject an instance of `NotificationStore` into the `NotificationHandlerService`.
- [x] Move logic from the `OpenNotificationAction` class into the handle method of the `NotificationHandler`.
- [x] Delete the file.
- [x] Remove references in the `NotificationsAssembler`.

orospakr:
Closes #332.

Depends on #350.


seanrucker (changes requested):
- [x] Delete the unnecessary Actions folder.
- [x] Remove the `actionProvider` property as it is no longer used.
- [x] Remove the `dispatcher` property as it is no longer used.
- [x] Move the `notification(for response: UNNotificationResponse)` method into an extension on `UNNotificationResponse`.
- [ ] Remove the `DispatchQueue` call to main thread as this should always be called on the main thread.
- [x] Make the `handle(_:completionHandler:)` method synchronous.

seanrucker (changes requested):
- [x] Remove the `get` sub definition since the property is read-only.
- [x] Compress the do/catch.
- [x] `openURL` is still dispatching to the main queue.
- [x] Your extension has incorrect header comment.

seanrucker (approved):

PR #356, opened 2018-11-15
Feature/330 simplify router

### LINKED ISSUE #330: Simplify Router
The purpose of the `Router` is to map URLs ("Deep Links" and Universal Links) to "actions" – some package of executable code. 

This was designed to be flexible and powerful – a deep/universal link can in theory launch any form of behaviour, whether it be UI-related or "headless" (to use Andrew's term). In-practice there are only three "routes" and the action each one invokes does the same thing – presents a view controller.

- presentSettings
- presentExperience?campaignID=123
   - or presentExperience?experienceID=123
   - or presentExperience?campaignURL=https://my-app.rover.io/abc123
- presentNotificationCenter

Further investigation reveals that the entire `Dispatcher`/`Action` mechanism is unnecessary and makes the code more difficult to reason about. This issue, simplifying the router, is the first step in removing the `Dispatcher` completely.

The simplified router's purpose can be rephrased:

The purpose of the router is to map URLs ("Deep Links" and Universal Links) to view controllers.

# Todo
- [x] Convert the `RouterService` to a `final class Router` and delete the `Router` protocol. The class is simply mapping three URLs to view controllers and everything needed to customize the behaviour is available through the assembler arguments and DI framework. The separate public protocol and private implementation adds unnecessary overhead.
- [x] The Router should have two similar methods for matching URLs: `viewController(for url: URL) -> UIViewController?` and `viewController(for userActivity: NSUserActivity) -> UIViewController?`. The former is used when the app delegate's `application(_:open:options)` method is called and the latter is used when the app delegate's `application(_:continue:restorationHandler:)` method is called. Both boil down to a URL (see current implementation). Note, the `handle` methods of the router are being removed. The router is only responsible for _mapping_ URLs to view controllers, _not_ for presenting them.
- [x] The router must first check the URL to determine if the scheme matches a Rover custom scheme or the domain matches a Rover associated domain. These values are passed into the constructor (see current implementation). If not a match, return nil. Assuming a match, the path of the URL and URL Query items must be checked next to verify this is a valid Rover URL, otherwise return nil. The logic for matching the URLs should be extracted from the three route handlers below and the route handlers should then be deleted. You will also need to remove the references to the route handlers from the various assemblers.
   - ExperienceRouteHandler
   - SettingsRouteHandler
   - NotificationCenterRouteHandler
- [x] Assuming a valid Rover URL, use the provider pattern to construct a corresponding view controller. The Router must take three providers in its constructor for this purpose:
   - `experienceViewControllerProvider: (ExperienceIdentifier) -> UIViewController`
   - `settingsViewControllerProvider: () -> UIViewController`
   - `notificationCenterViewControllerProvider: () -> UIViewController`
- [x] All three routes currently have the prefix "present" which implies a verb as opposed to a noun and is redundant. Support the same route structure minus the "present' prefix, but support  the current version as well for backwards compatibility. For example, both of the following URLs would map to the settings view controller:
   - rv-my-app://settings
   - rv-my-app://presentSettings
- [x] The `NotificationCenterViewController` uses router when a notification is tapped with an `openURL` tap behavior. Modify the code to ask the router for a view controller instead of using a provider. If a view controller is obtained, set the transitioning delegate to the `` and present it manually. If a view controller is not obtained just call `UIApplication.open(_:options:completionHandler:)`.

orospakr:
Resolves #330.


seanrucker (changes requested):
- [x] Remove warning (unused dispatcher constant)
- [ ] Sort out the unused `isUniversalLink` function
- [x] Please test alternate valid version of an experience deep link as well as an invalid (incorrect url params).

seanrucker (changes requested):
- [x] Please read through the `isUniversalLink` function, understand it and use appropriately.

seanrucker (changes requested):
- [ ] Restore InfluenceTracker behaviour
- [ ] Comment in Router file

seanrucker (approved):

PR #360, opened 2018-11-16
Feature/create sync target

orospakr:


seanrucker (changes requested):
- [x] The `accountToken` and `endpoint` arguments need be moved from `DataAssembler` to the `SyncAssembler`.
- [x] The moved files all need their header comments updated.
- [x] RoverSync.h needs to be stripped down, take a look at RoverUI.h

seanrucker (approved):

PR #361, opened 2018-11-20
Ripout Event/EventQueue.

### LINKED ISSUE #354: Remove EventQueue from RoverData
- [x] Delete the entire EventQueue folder from the RoverData module, save for the `EventInfo` struct. Move `EventInfo` into the root folder of the module for now.
- [x] Delete event-related entries from the `DataAssembler`:
   - EventsClient
   - EventQueue
- [x] Delete the `containerDidAssemble(resolver:)` method from `DataAssembler`.

orospakr:
Closes #354.


seanrucker (changes requested):
Need to remove all the constructor arguments from the DataAssembler as they are all event-related and no longer part of this module. Some of them will be re-added to the Sync module at a later date but its not worth moving them (vs deleting) as they are just simple properties.

seanrucker (changes requested):
You've removed the public constructor completely which means you can no longer instantiate the DataAssembler from your AppDelegate. You need to create a public empty constructor. Swift creates a default constructor for structs, however the access is internal by default. You need to opt-in to make a struct publicly instantiateable (<- is that a word?).

seanrucker (approved):

PR #362, opened 2018-11-20
Feature/341:  Rename Context to DeviceSnapshot, collapse Content Providers, migrate to UI module.

### LINKED ISSUE #341: Move Context from RoverData to RoverUI
The word Context becomes confusing when we're using Core Data. We use the word Device elsewhere in our stack which makes it a reasonable replacement name. In addition, our model for capturing context, persisting some values and letting other modules contribute to the process is unnecessarily complicated. 

Also, most of the computed properties on the Context require UIKit so this class needs to be moved out of the RoverData module and into the RoverUI module.

## Todo
- [x] Rename the `Context` struct to `DeviceSnapshot` and _keep_ it in the RoverData module (put it in the root of the module for now).
- [x] Add a default value of `nil` to all initializer arguments so the `DeviceSnapshot` can be initialized without providing any arguments. I.e. `let snapshot = DeviceSnapshot()`.
- [x] Replace the following files with a single open class named `Device` which uses computed properties to return values. As an open class with computed properties, this allows easy developer customization with very little overhead. For the persisted values, use `UserDefaults` directly instead of the `PersistedValue` utility class. Save the `Device` class in the RoverUI module, not RoverData, and ensure the target membership is correct.
   - ContextManager
   - ContextProvider
   - ModularContextProvider
   - TokenManager
   - UserInfo
   - DebugContextProvider
   - LocaleContextProvider
   - NotificationsContextProvider
   - PushTokenContextProvider
   - ReachabilityContextProvider
   - StaticContextProvider
   - TimeZoneContextProvider
   - UserInfoContextProvider
- [x] For each of the 4 remaining ContextProviders, rename the protocols from "ContextProvider" to "InfoProvider".
   - AdSupportInfoProvider
   - BluetoothInfoProvider
   - LocationInfoProvider
   - TelephonyInfoProvider
- [x] The `Device` class should accept optional instances of the 4 info providers in its constructor and the `Device` class should delegate to these in its computed properties as it does today.
- [x] Add a computed variable named `snapshot` which builds and returns an instance of the `DeviceSnapshot` _struct_ from all the computed properties on the `Device` _class_.
- [x] Register the `Device` class in the `UIAssembler`.
- [x] Move the vendored `Reachabilty` class into RoverUI.
- [x] Delete the Context folder from RoverData.
- [x] Delete all context-related entries from the `DataAssembler`:
   - ContextManager
   - ContextProvider
   - LocaleContextProvider
   - PushTokenContextProvider
   - ReachabilityContextProvider
   - StaticContextProvider
   - TimeZoneContextProvider
   - TokenManager
   - UserInfoManager
   - UserInfoContextProvider

orospakr:
Closes #341.


Rover.podspec
```diff
@@ -43,29 +43,34 @@ Pod::Spec.new do |s|
     ss.source_files = "Sources/Location/**/*.swift"
     ss.resources = "Sources/Location/Model/RoverLocation.xcdatamodeld"
     ss.dependency "Rover/Data"
```
seanrucker: RoverData is a transient dependency of RoverUI so you can remove this dependency if the subspec depends on RoverUI.

seanrucker: Same for all other subspacs that depend on RoverUI now.

Sources/UI/Device/Device.swift
```diff
@@ -0,0 +1,405 @@
...
+//
+
+import Foundation
+import os
+import UIKit
+import UserNotifications
+
+class Device {
+    
+    let userDefaults: UserDefaults
```
seanrucker: Can we remove the injected reference to `UserDefaults`, `JSONEncoder` and `JSONDecoder`? I've found the explicit rule of "every dependency must be injected" only adds to complexity and provides zero value. E.g. there is no reason to ever use any instance of `UserDefaults` other than `UserDefaults.standard`. Singletons have their drawbacks but it is the "iOS way" as Apple uses it _everywhere_.

Sources/UI/Device/Device.swift
```diff
@@ -0,0 +1,405 @@
...
+    let jsonDecoder: JSONDecoder
+    let jsonEncoder: JSONEncoder
+    
+    let adSupportInfoProvider: AdSupportInfoProvider?
+    let bluetoothInfoProvider: BluetoothInfoProvider?
+    let telephonyInfoProvider: TelephonyInfoProvider?
+    let locationInfoProvider: LocationInfoProvider?
+    
+    let reachability = Reachability(hostname: "google.com")!
+    let userNotificationCenter = UNUserNotificationCenter.current()
```
seanrucker: For same reasons as unnecessary injections, my preference is to refer to `UNUserNotificationCenter.current()` in the function where you actually need to use it instead of creating a constant on the class. It adds indirection as the user looking at the function implementation only knows he's dealing with an instance property named `userNotificationCenter` and has to follow the source to realize what you've done. If you create the instance in-line where you are using it, it adds clarity because developers are very accustomed to seeing `UNUserNotificationCenter.current()` and immediately understand, "ok you're accessing the UNUserNotifiicationCenter` singleton which is standard practice".

Sources/UI/Device/Device.swift
```diff
@@ -0,0 +1,405 @@
...
+        self.jsonEncoder = jsonEncoder
+        self.adSupportInfoProvider = adSupportInfoProvider
+        self.bluetoothInfoProvider = bluetoothInfoProvider
+        self.telephonyInfoProvider = telephonyInfoProvider
+        self.locationInfoProvider = locationInfoProvider
+    }
+    
+    public var snapshot: DeviceSnapshot {
+        return DeviceSnapshot(
+            advertisingIdentifier: advertisingIdentifier,
```
seanrucker: Can you refer to all of these with `self`? I used to omit the `self` keyword but have found it adds clarity and reduces bugs. It is immediately obvious to developers "ah this property is defined as an instance variable on the class" instead of "where did this `advertisingIdentifer` variable come from? Is it a function variable or a instance property of the class?".

Sources/UI/Device/Device.swift
```diff
@@ -0,0 +1,405 @@
...
+            sdkVersion: sdkVersion,
+            carrierName: carrierName,
+            radio: radio,
+            isTestDevice: isTestDevice,
+            timeZone: timeZone,
+            userInfo: userInfo
+        )
+    }
+    
+    public private(set) var pushToken: DeviceSnapshot.PushToken? {
```
seanrucker: This, along with all other computed properties, needs to be `open` as opposed to `public` so the developer can customize the implementation.

Sources/UI/Device/Device.swift
```diff
@@ -0,0 +1,405 @@
...
+//  Created by Andrew Clunis on 2018-11-20.
+//  Copyright © 2018 Rover Labs Inc. All rights reserved.
+//
+
+import Foundation
+import os
+import UIKit
+import UserNotifications
+
+class Device {
```
seanrucker: Needs to be an `open` class so the developer can customize the implementation.

Sources/UI/Device/Device.swift
```diff
@@ -0,0 +1,405 @@
...
+    
+    public var localeRegion: String? {
+        return Locale.current.regionCode
+    }
+    
+    public var localeScript: String? {
+        return Locale.current.scriptCode
+    }
+    
+    // MARK: Locale
```
seanrucker: Incorrect `MARK` label.

Sources/UI/Device/Device.swift
seanrucker: If you were to apply this pattern consistently in this class, you'd need to inject a ton more instances to avoid using their singletons. E.g. `UIDevice.current`, `Locale.current`, `Bundle.main` etc. It makes the constructor and assembler unnecessarily complex and doesn't add any value or additional configurability as each computed property can be overridden if needed.

Sources/UI/Device/Device.swift
```diff
@@ -0,0 +1,405 @@
...
+            } catch {
+                os_log("Failed to encode user info: %@", log: .general, type: .error, error.localizedDescription)
+            }
+        }
+    }
+    
+    // MARK: Notifications Authorization
+    
+    public var notificationAuthorization: String {
+        // Refresh status for _next_ time context is requested
```
seanrucker: Update comment since we're moving away from "context". Suggested re-wording: "Refresh status next time snapshot is created".

seanrucker (changes requested):
In addition to inline comments:
- Maybe move `DeviceModel` into the `Device` folder?
- `DeviceModel` has incorrect header comment.

Rover.podspec
orospakr: gonna review your review! "transitive" ;)

seanrucker (approved):

PR #368, opened 2018-12-10
Make region monitoring limits configurable.

### LINKED ISSUE #367: Configure max beacon/geofence regions to monitor
Add two configuration params to the `LocationAssembler` that are used to determine how many wildcard regions to monitor for and how many geofence regions to monitor for. Here's some rules for valid values:

- `maxGeofenceRegionsToMonitor` must be less than or equal to 20.
- The default value of `maxGeofencesToMonitor` is 20.
- `maxBeaconRegionsToMonitor` must be less than or equal to `maxGeofenceRegionsToMonitor`.
- The default value of `maxBeaconRegionsToMonitor ` is 5.

The `LocationManager` first monitors for _up to_ `maxBeaconRegionsToMonitor` wildcard regions, then monitors for `maxGeofenceRegionsToMonitor` minus the number of wildcard regions it is monitoring for.

orospakr:
Closes #367.

### DIFF
--- a/Sources/Location/LocationManager.swift
+++ b/Sources/Location/LocationManager.swift
@@ -44,10 +44,10 @@ class LocationManager {
         self.maxBeaconRegionsToMonitor = maxBeaconRegionsToMonitor
         let theoreticalMaximumGeofences = 20
         if maxGeofenceRegionsToMonitor > theoreticalMaximumGeofences {
-            assertionFailure("You may only specify that Rover can monitor up to \(theoreticalMaximumGeofences) geofences at a time.")
+            fatalError("You may only specify that Rover can monitor up to \(theoreticalMaximumGeofences) geofences at a time.")
         }
         if maxBeaconRegionsToMonitor >= maxGeofenceRegionsToMonitor {
-            assertionFailure("Rover uses the same region slots for monitoring beacons as it does for monitoring geofences, so therefore you may only specify that Rover can monitor up to the same max number of beacons as geofence regions, currently \(maxGeofenceRegionsToMonitor).")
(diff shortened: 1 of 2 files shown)


Sources/Location/LocationManager.swift
```diff
@@ -31,7 +34,21 @@ class LocationManager {
...
+        context: NSManagedObjectContext,
+        eventQueue: EventQueue,
+        maxGeofenceRegionsToMonitor: Int,
+        maxBeaconRegionsToMonitor: Int
+    ) {
+        self.maxGeofenceRegionsToMonitor = maxGeofenceRegionsToMonitor
+        self.maxBeaconRegionsToMonitor = maxBeaconRegionsToMonitor
+        let theoreticalMaximumGeofences = 20
+        if maxGeofenceRegionsToMonitor > theoreticalMaximumGeofences {
+            assertionFailure("You may only specify that Rover can monitor up to \(theoreticalMaximumGeofences) geofences at a time.")
```
seanrucker: In this case I would say it's more appropriate to use `fatalError` since the developer is _guaranteed_ to execute this code path during development and we can't really recover from this.

Sources/Location/LocationManager.swift
```diff
@@ -31,7 +34,21 @@ class LocationManager {
...
+        }
+        if maxBeaconRegionsToMonitor >= maxGeofenceRegionsToMonitor {
+            assertionFailure("Rover uses the same region slots for monitoring beacons as it does for monitoring geofences, so therefore you may only specify that Rover can monitor up to the same max number of beacons as geofence regions, currently \(maxGeofenceRegionsToMonitor).")
```
seanrucker: Use `fatalError` here as well.

seanrucker (changes requested):

seanrucker (approved):

PR #369, opened 2018-12-10
NotificationExtensionHelper now gives feedback if Rover handled the notification.

### LINKED ISSUE #359: NotificationExtensionHelper should provide feedback on whether the push is a Rover push or not

orospakr:
Closes #359.

### DIFF
--- a/Sources/AppExtensions/NotificationExtensionHelper.swift
+++ b/Sources/AppExtensions/NotificationExtensionHelper.swift
@@ -19,10 +19,10 @@ public class NotificationExtensionHelper {
         self.userDefaults = userDefaults
     }
     
-    public func didReceive(_ request: UNNotificationRequest, withContent content: UNMutableNotificationContent) {
+    public func didReceive(_ request: UNNotificationRequest, withContent content: UNMutableNotificationContent) -> Bool {
         guard let data = try? JSONSerialization.data(withJSONObject: content.userInfo, options: []) else {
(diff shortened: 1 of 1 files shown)


Sources/AppExtensions/NotificationExtensionHelper.swift
```diff
@@ -19,10 +19,10 @@ public class NotificationExtensionHelper {
         self.userDefaults = userDefaults
     }
     
-    public func didReceive(_ request: UNNotificationRequest, withContent content: UNMutableNotificationContent) {
+    public func didReceive(_ request: UNNotificationRequest, withContent content: UNMutableNotificationContent) -> Bool {
```
seanrucker: Can you add `@discardableResult` to this function? This allows developers to use this function without _requiring_ them to pay attention to the return value. Unless the app uses additional push providers, the developer will likely not care about the return value. Also, without adding `@discardableResult` Xcode would generate warnings for developers updating to 2.3.0 from a previous version which could be considered a "breaking change".

seanrucker (changes requested):

Sources/AppExtensions/NotificationExtensionHelper.swift
orospakr: Oho. I was wondering how to silence that warning. You got it.

seanrucker (approved):

PR #372, opened 2018-12-10
Feature/328 events in core data

### LINKED ISSUE #328: Event Model
- [x] Add a new entity to the .xcdatamodeld file named "Event". Add the following list of attributes, making sure the optional checkbox is _unchecked_ for any of the non-optional values:
   - id: Transformable
   - name: String
   - namespace: String?
   - attributes: Transformable
   - deviceSnapshot: Transformable
   - timestamp: Date
   - isFlushed: Bool (set default to `false`)
- [x] Create a corresponding `Event` class that extends `NSManagedObject`.  Look at the `Beacon` and `Geofence` classes for how to structure this. Create the corresponding `@NSManaged` properties with the following corresponding types:
   - id: UUID
   - name: String
   - namespace: String?
   - attributes: Attributes
   - deviceSnapshot: DeviceSnapshot
   - timestamp: Date
   - isFlushed: Bool
- [x] Override the `awakeFromInsert()` method on the `Event`. In the implementation call `super.awakeFromInsert()` and then set default values for id, attributes, device and timestamp.
   - `self.id = UUID()`
   - `self.attributes = Attributes()`
   - `self.device = DeviceSnapshot()`
   - `self.timestamp = Date()`
- [x] Use the technique outlined in the Core Data book to register custom value transformers for the `deviceSnapshot` and `attributes` properties.

orospakr:
Closes #328.


Sources/Data/Model/Event.swift
```diff
@@ -0,0 +1,36 @@
...
+/// These are used both for store-and-forwarding events to the Rover cloud services, but also kept and queried locally to power automated campaigns.
+final class Event : NSManagedObject {
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<Event> {
+        return NSFetchRequest<Event>(entityName: "Event")
+    }
+    
+    @NSManaged internal(set) var id: String
+    @NSManaged internal(set) var name: String
+    @NSManaged internal(set) var namespace: String?
+    @NSManaged internal(set) var attributes: Attributes?
```
seanrucker: Does this need to be optional? It has a default value set in the `awakeFromInsert` so the optionality here seems erroneous.

Sources/Data/EventInfo.swift
```diff
@@ -11,10 +11,26 @@ import Foundation
...
+    ///
+    /// * `String`
+    /// * `Int`
+    /// * `Double`
+    /// * `Bool`
+    /// * `[String]`
+    /// * `[Int]`
+    /// * `[Double]`
+    /// * `[Bool]`
+    /// * `[String: Any]` (but the `Any` here may not be another dictionary.)
```
seanrucker: I think there is a slight misunderstanding of the types. The dictionary value is of `String` to `Any` where `Any` has the same constraints listed above. I.e it _can_ be another nested dictionary.

Sources/Data/Attributes.swift
```diff
@@ -0,0 +1,299 @@
...
+//  RoverData
+//
+//  Created by Sean Rucker on 2018-07-24.
+//  Copyright © 2018 Rover Labs Inc. All rights reserved.
+//
+
+import Foundation
+import os
+
+/// Wraps a [String: Any], but enables use of both NSCoding and Codable.  Enforces the
```
seanrucker: Looks like you didn't finish your thought in this comment.

Sources/Data/Attributes.swift
```diff
@@ -0,0 +1,299 @@
...
+class Attributes: NSObject, NSCoding, Codable {
+    var rawValue: [String: Any]
+    
+    public init(_ dictionary: [String:Any]) {
```
seanrucker: If you rename this initializer to `public init(rawValue: [String: Any])` you can conform to `RawValueRepresentable`.

Sources/Data/Attributes.swift
```diff
@@ -0,0 +1,299 @@
...
+                    return
+                }
+                
+                if let value = try? container.decode(String.self, forKey: key) {
+                    assembledHash[keyString] = value
+                    return
+                }
+                
+                if (try? container.decodeNil(forKey: key)) == true {
+                    assembledHash[keyString] = nil
```
seanrucker: [String: Any] can't contain `nil` values. Setting the value to `nil` simply omits it from the Dictionary and is therefore a no-op.

Sources/Data/Attributes.swift
```diff
@@ -0,0 +1,299 @@
...
+                }
+                
+                // now try probing for an embedded dict.
+                if let dictionary = try? container.nestedContainer(keyedBy: DynamicCodingKeys.self, forKey: key) {
+                    assembledHash[keyString] = try fromKeyedDecoder(dictionary)
+                    return
+                }
+                
+                // now try probing for an array.
+                if var array = try? container.nestedUnkeyedContainer(forKey: key) {
```
seanrucker: I believe this section can be replaced with simple decode calls that follow the pattern above. E.g.:
```swift
if let value = try? container.decode([String].self, forKey: key) {
    ...
}
```

Sources/Data/Attributes.swift
```diff
@@ -0,0 +1,299 @@
...
+                switch value {
+                case let value as Int:
+                    try container.encode(value, forKey: key)
+                case let value as Bool:
+                    try container.encode(value, forKey: key)
+                case let value as String:
+                    try container.encode(value, forKey: key)
+                case let value as Double:
+                    try container.encode(value, forKey: key)
+                case let array as [Any]:
```
seanrucker: Similar to the comment made in decoder, you can replace this with simple casting cases. E.g.:
```swift
case let value as [String]:
    ...
case let value as [Double]:
    ...
```

Sources/Data/Attributes.swift
```diff
@@ -0,0 +1,299 @@
...
+        rawValue = [:]
+        super.init()
+    }
+}
+
+
+typealias AttributeValue = [String: Any]
+
+extension Dictionary where Key == String, Value: Any {
+    fileprivate var asAttributesForNsDictionary: [String: Any] {
```
seanrucker: If you move this into the `Attributes` initializer, you will only need to reference this in one place. The `var attributes: NSDictionary` can be moved into the `func encode(with aCoder: NSCoder)` at which point you know you the `rawValue` property is guaranteed to already contain a valid dictionary.

Sources/Data/Attributes.swift
```diff
@@ -0,0 +1,299 @@
...
+            default:
+                guard let transformed = transformAsNeeded(value: element.value) else {
+                    return
+                }
+                result[element.key] = transformed
+            }
+        }
+    }
+    
+    var attributes: NSDictionary {
```
seanrucker: As mentioned above, if you move this into the `func encode(with aCoder: NSCoder)` and implement the above suggestion you no longer need to call `self.asAttributesForNsDictionary` as that will have been handled in the initializer and you can work with `rawValue` directly as it is guaranteed to contain a valid dictionary.

Sources/Data/Attributes.swift
```diff
@@ -0,0 +1,299 @@
...
+        let dictionary = NSMutableDictionary()
+        // setValuesForKeys coerces Swifty primitives to their NS* equivalents, however, in asAttributesForNsDictionary() we do a few extra transforms.
+        // setValuesForKeys' equivalent in the opposite direction is NSDictionary.dictionaryWithValues
+        let attributesSuitableForNsDictionary = self.asAttributesForNsDictionary
+        dictionary.setValuesForKeys(attributesSuitableForNsDictionary)
+        return dictionary
+    }
+    
+    /// Undo the effects of [asAttributesForNsDictionary].
+    fileprivate var fromAttributesAsNsDictionary: [String: Any] {
```
seanrucker: This, along with the logic in the `NSDictionary` extension, can be moved directly into the `required init?(coder aDecoder: NSCoder)` function.

Sources/Data/Attributes.swift
```diff
@@ -0,0 +1,299 @@
+//
```
seanrucker, line 1: In general, I found the code difficult to reason about due to the mixing of inline code logic and private extensions. I've added review suggestions throughout the code. Here's a summary of how this could be simplified:
 
 1. Initializer accepts a [String: Any] and validates all keys/values comply with our restrictions. Mark this initializer with `throws` instead of using `assertionFailure` so we can move the error handling strategy upstream.
 2. The NSCoding encode function wraps the already validated rawValue property into an NSDictionary and encodes it.
 3. The NSCoding decode function decodes an NSDictionary, converts it to [String: Any] and calls the initializer mentioned above.
 4. The Codable encode function encodes the already validated rawValue property into by using a casting pattern match in a switch statement.
 5. The Codable decode function decodes a [String: Any] type by attempting to decode each of the known valid types, then finally calls the initializer mentioned above to guarantee the `rawValue` property is valid.

Sources/Data/DeviceSnapshot.swift
```diff
@@ -241,3 +213,186 @@ public struct DeviceSnapshot: Codable, Equatable {
         self.userInfo = userInfo
     }
 }
+
+// MARK: Push Token
+
+class DevicePushToken: NSObject, NSCoding, Codable {
```
seanrucker: Suggested alternative naming strategy for these classes: Append `Snapshot` instead of prepending `Device`. I.e.

- DeviceSnapshot
- PushTokenSnapshot
- LocationSnapshot
- etc.

Instead of:

- DeviceSnapshot
- DevicePushToken
- DeviceLocation
- etc.

Just my opinion, but I feel the `Snapshot` approach better "groups" these objects together and better describes  their intended purpose.

Sources/Location/Extensions/Location.CLBeaconRegion.swift
```diff
@@ -9,7 +9,7 @@
 import CoreLocation
 
 extension CLBeaconRegion: AttributeRepresentable {
```
seanrucker: `AttributeRepresentable` has been removed in this PR. You'll need to update this to match the approach you took in the `Beacon` and `Geofence` classes.

Sources/Location/Extensions/Location.CLCircularRegion.swift
```diff
@@ -9,7 +9,7 @@
 import CoreLocation
 
 extension CLCircularRegion: AttributeRepresentable {
```
seanrucker: `AttributeRepresentable` has been removed in this PR. You'll need to update this to match the approach you took in the `Beacon` and `Geofence` classes.

Sources/UI/Experiences/Models/Block.swift
```diff
@@ -20,7 +20,7 @@ public protocol Block: AttributeRepresentable, Decodable {
 }
 
 extension Block {
```
seanrucker, line 22: `AttributeRepresentable` has been removed in this PR. You'll need to update this to match the approach you took in the `Beacon` and `Geofence` classes.

Sources/UI/Experiences/Models/BlockTapBehavior.swift
```diff
@@ -18,7 +18,7 @@ public enum BlockTapBehavior {
 // MARK: AttributeRepresentable
 
 extension BlockTapBehavior: AttributeRepresentable {
```
seanrucker: `AttributeRepresentable` has been removed in this PR. You'll need to update this to match the approach you took in the `Beacon` and `Geofence` classes.

Sources/UI/Experiences/Models/Experience.swift
```diff
@@ -61,7 +61,7 @@ extension Experience: Decodable {
 // MARK: AttributeRepresentable
 
 extension Experience: AttributeRepresentable {
```
seanrucker: `AttributeRepresentable` has been removed in this PR. You'll need to update this to match the approach you took in the `Beacon` and `Geofence` classes.

Sources/UI/Experiences/Models/Row.swift
```diff
@@ -108,7 +108,7 @@ extension Row: Decodable {
 }
 
 extension Row: AttributeRepresentable {
```
seanrucker: `AttributeRepresentable` has been removed in this PR. You'll need to update this to match the approach you took in the `Beacon` and `Geofence` classes.

Sources/UI/Experiences/Models/Screen.swift
```diff
@@ -72,7 +72,7 @@ public struct Screen: Decodable {
 // MARK: AttributeRepresentable
 
 extension Screen: AttributeRepresentable {
```
seanrucker: `AttributeRepresentable` has been removed in this PR. You'll need to update this to match the approach you took in the `Beacon` and `Geofence` classes.

Sources/UI/Experiences/UI/ScreenViewController.swift
```diff
@@ -99,8 +99,8 @@ open class ScreenViewController: UICollectionViewController, UICollectionViewDat
     override open func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
         
-        let attributes: Attributes = [
```
seanrucker: Should be of type `[String: Any]` as opposed to `Attributes`.

Sources/UI/Notifications/Models/Notification.swift
```diff
@@ -51,7 +51,7 @@ extension Notification {
 }
 
 extension Notification: AttributeRepresentable {
```
seanrucker: `AttributeRepresentable` has been removed in this PR. You'll need to update this to match the approach you took in the `Beacon` and `Geofence` classes.

Sources/UI/Notifications/Services/InfluenceTracker/InfluenceTrackerService.swift
```diff
@@ -59,7 +59,7 @@ class InfluenceTrackerService: InfluenceTracker {
             var campaignID: ID
             var receivedAt: Date
             
-            var attributeValue: AttributeValue {
```
seanrucker, line 62: `AttributeRepresentable` has been removed in this PR. You'll need to update this to match the approach you took in the `Beacon` and `Geofence` classes.

seanrucker (changes requested):
In addition to all the inline comments, you should replace all instances of ID with String since it was removed in this PR.

Sources/Data/Attributes.swift
orospakr: Worked nicely. Good idea.

Sources/Data/Attributes.swift
orospakr: Likewise.

seanrucker (approved):

PR #373, opened 2018-12-17
Feature/353 events pipeline service

### LINKED ISSUE #353: EventPipeline Service
The `EventPipeline` is responsible for creating/tracking events and notifying observers when events have been added. With a reference to the `NSManagedObjectContext` the `EventPipline` provides a simple API to add an event. It also wraps the `NSManagedObjectContextObjectsDidChange` API on the `NSManagedObjectContext` and provides a simple API to be notified when an event has been added.

## Todo

- [x] Create a public, open `EventPipeline` class. The initializer accepts an instance of `NSManagedObjectContext` as well as an instance of the `Device` class.
- [x] Register the `EventPipeline` class as a singleton in the DataAssembler. Resolved an instance of the `NSManagedObjectContext` named "background" to pass in the constructor.
- [x] Create an `addEvent(_:)` method which accepts an instance of `EventInfo`. The implementation should insert an `Event` object into the context, set the properties based on the values of the `EventInfo` and save the context. (The event.deviceSnapshot is calculated from the `Device` class injected through  the `EventPipeline` constructor. This should happen from within the `context.perform` block. 
- [x] In the event of an error in `addEvent(_:)` the context should be rolled back and the error logged. See the `insertObjects(from:)` method of the `PagingSyncParticipant` for how to check for and log multiple errors.
- [x] Add a public `ObserverSet` property to the `EventPipeline`: `var observers = ObserverSet<Event>()`.
- [x] Add a property to the `EventPipeline` class to reference an observer token: `var objectsDidChangeObserver: NSObject!`
- [x] In the initializer, where the `NSManagedObjectContext` is passed, add an observer: 
```
NotificationCenter.default.addObserver(forName: .NSManagedObjectContextObjectsDidChange, object: context, queue: nil) { [weak self] notification in
     
}
```
- [x] In the `deinit` remove the observer: `NotificationCenter.default.removeObserver(objectsDidChangeObserver)`.
- [x] In the implementation of the observer block, obtain the list of inserted objects from the `NSInsertedObjectsKey` key. Filter the set for `Event` objects since that's all we care about. If the set is not empty, notify the observers in the `ObserverSet`, one at a time for each updated event.

orospakr:
Closes #353.

### DIFF
--- a/Sources/Data/EventPipeline.swift
+++ b/Sources/Data/EventPipeline.swift
@@ -31,11 +31,7 @@ public class EventPipeline {
             let insertedEvents = insertedObjects
                 .compactMap({ $0 as? Event })
 
-            insertedEvents.forEach({ self?.observers.notify(parameters: $0) })
-            
-            insertedEvents.forEach({ (event) in
-                self?.observers.notify(parameters: event)
-            })
+            insertedEvents.forEach({ event in self?.observers.notify(parameters: event) })
         }
     }
     
--- a/Sources/Data/EventPipeline.swift
+++ b/Sources/Data/EventPipeline.swift
@@ -14,23 +14,24 @@ class EventPipeline {
     public var observers = ObserverSet<Event>()
     
     private let managedObjectContext: NSManagedObjectContext
+    private let deviceInfoProvider: DeviceInfoProvider
     private var objectsDidChangeObserver: NSObjectProtocol!
     
-    init(managedObjectContext: NSManagedObjectContext) {
+    init(
+        managedObjectContext: NSManagedObjectContext,
+        deviceInfoProvider: DeviceInfoProvider
+    ) {
         self.managedObjectContext = managedObjectContext
-        
+        self.deviceInfoProvider = deviceInfoProvider
         self.objectsDidChangeObserver = NotificationCenter.default.addObserver(forName: .NSManagedObjectContextObjectsDidChange, object: managedObjectContext, queue: nil) { [weak self] notification in
             guard let insertedObjects = notification.userInfo?[NSInsertedObjectsKey] as? Set<NSObject> else {
                 return
             }
             
-            insertedObjects
+            let insertedEvents = insertedObjects
                 .compactMap({ $0 as? Event })
-                .forEach({ self?.observers.notify(parameters: $0) })
-            
-//            let insertedEvents = insertedObjects.filter({ (potentialInsertedEvent) -> Bool in
-//                return potentialInsertedEvent is Event
-//            }) as! Set<Event>
+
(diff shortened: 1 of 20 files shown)


Sources/Data/EventPipeline.swift
```diff
@@ -0,0 +1,80 @@
...
+            guard let insertedObjects = notification.userInfo?[NSInsertedObjectsKey] as? Set<NSObject> else {
+                return
+            }
+            
+            let insertedEvents = insertedObjects
+                .compactMap({ $0 as? Event })
+
+            insertedEvents.forEach({ self?.observers.notify(parameters: $0) })
+            
+            insertedEvents.forEach({ (event) in
```
seanrucker: This seems to be a dupe of the previous line.

Sources/Data/EventPipeline.swift
```diff
@@ -0,0 +1,80 @@
...
+        event.deviceSnapshot = deviceSnapshot
+        event.name = eventInfo.name
+        event.namespace = eventInfo.namespace
+        
+        managedObjectContext.perform { [managedObjectContext] in
+            managedObjectContext.insert(event)
+           
+            do {
+                try managedObjectContext.save()
+                managedObjectContext.reset()
```
seanrucker: Copy pasta?

Sources/Data/Model/Event.swift
```diff
@@ -26,11 +26,13 @@ final class Event : NSManagedObject {
...
     
-    override func awakeFromInsert() {
+    public override func awakeFromInsert() {
         // default values for newly created and inserted Events.
+        
         self.id = UUID().uuidString
         self.attributes = Attributes()
         self.deviceSnapshot = DeviceSnapshot()
         self.timestamp = Date()
+        self.isFlushed = false
```
seanrucker: This is a no-op because your xcdatamodeld file already sets the default to "NO" (false).

Sources/Location/LocationManager.swift
```diff
@@ -12,7 +12,9 @@ import os.log
 
 class LocationManager {
     let context: NSManagedObjectContext
-    let eventQueue: EventQueue
+    
+    // Now there's a cyclic dependency in DI through here. D'oh!
```
seanrucker: Can you remove this comment?

Sources/UI/UIAssembler.swift
```diff
@@ -7,6 +7,7 @@
 //
 
 import SafariServices
+import CoreData
```
seanrucker: Erroneous?

seanrucker (changes requested):

seanrucker (approved):

PR #379, opened 2018-12-17
Feature/377 create campaigns module

orospakr:


Sources/UI/Notifications/Services/NotificationStore/NotificationStoreService.swift
```diff
@@ -1,166 +1,166 @@
...
+//
+//        guard let eventPipeline = self.eventPipeline else {
+//            return
+//        }
+//
+//        let attributes: Attributes = ["notification": notification]
+//        let event = EventInfo(name: "Notification Marked Read", namespace: "rover", attributes: attributes)
+//        eventPipeline.addEvent(event)
+//    }
+//}
```
orospakr, line 166: This remains only just for temporary reference, I'll be replacing it next PR with Core Data.

seanrucker: 👍

Sources/UI/UIAssembler.swift
```diff
@@ -17,21 +17,12 @@ public struct UIAssembler {
     public var isLifeCycleTrackingEnabled: Bool
     public var isVersionTrackingEnabled: Bool
     
-    public var appGroup: String?
-    public var influenceTime: Int
-    public var isInfluenceTrackingEnabled: Bool
-    public var maxNotifications: Int
-    
-    public init(associatedDomains: [String] = [], urlSchemes: [String] = [], sessionKeepAliveTime: Int = 30, isLifeCycleTrackingEnabled: Bool = true, isVersionTrackingEnabled: Bool = true, appGroup: String? = nil, isInfluenceTrackingEnabled: Bool = true, influenceTime: Int = 120, maxNotifications: Int = 200) {
+    public init(associatedDomains: [String] = [], urlSchemes: [String] = [], sessionKeepAliveTime: Int = 30, isLifeCycleTrackingEnabled: Bool = true, isVersionTrackingEnabled: Bool = true, appGroup: String? = nil, influenceTime: Int = 120) {
```
seanrucker: Should `appGroup` and `influenceTime` still be in there?

Sources/Campaigns/Extensions/UI.UNNotificationResponse.swift
```diff
@@ -10,14 +10,14 @@ import Foundation
 import UserNotifications
 
 extension UNNotificationResponse {
```
seanrucker: Comment should be `RoverCampaigns` and the filename should be `Campaigns.OSLog` not `UI.OSLog`.

seanrucker: Also looks like the target is wrong.

seanrucker (changes requested):

Sources/Campaigns/Extensions/UI.UNNotificationResponse.swift
seanrucker: The comment in the header is wrong.

seanrucker (changes requested):
The comment in the header of the Campaigns.UNNotificationResponse.swift file is wrong.

seanrucker (approved):

PR #380, opened 2018-12-20
Feature/374 notifications data model

orospakr:


Sources/Campaigns/Extensions/Campaigns.UNNotificationResponse.swift
```diff
@@ -10,19 +10,21 @@ import Foundation
 import UserNotifications
 
 extension UNNotificationResponse {
+    // TODO change to return a campaign.
```
orospakr, line 13: If even that.

Sources/Data/Model/Event.swift
```diff
@@ -35,3 +35,64 @@ public final class Event : NSManagedObject {
...
+            
+            if !insertedEvents.isEmpty {
+                observerCallback(insertedEvents)
+            }
+        }
+    }
+    
+    public func attemptInsert() {
+        guard let managedObjectContext = self.managedObjectContext else {
+            assertionFailure("Not associated with a managed object context, cannot delete.")
```
seanrucker: Should "cannot delete" be "cannot insert"?

Sources/Data/Model/Event.swift
```diff
@@ -35,3 +35,64 @@ public final class Event : NSManagedObject {
...
+                } else {
+                    os_log("Unable to save event into the EventPipeline. Dropping it. Reason: %s", log: .persistence, type: .error, (error as NSError).userInfo.debugDescription)
+                }
+                
+                managedObjectContext.rollback()
+            }
+        }
+    }
+    
+    public convenience init(
```
seanrucker, line 53: The word `create` in the first param name seems a bit out of convention. I think because its a constructor it already implies "create". Maybe change `createFrom` to just `from`.

Sources/Data/Model/Notification.swift
```diff
@@ -0,0 +1,139 @@
...
+    @NSManaged public var isRead: Bool
+    
+    public override func awakeFromInsert() {
+        self.id = UUID()
+        super.awakeFromInsert()
+    }
+    
+    // MARK: Codable
+    
+    private enum CodingKeys: String, CodingKey {
```
seanrucker: It doesn't look like this class conforms to `codable` so are these coding keys necessary?

Sources/Data/Model/Notification.swift
```diff
@@ -0,0 +1,139 @@
...
+extension Notification {
+    public func markRead() {
+        guard let managedObjectContext = self.managedObjectContext else {
+            assertionFailure("Not associated with a managed object context, cannot delete.")
+            return
+        }
+        
+        self.isRead = true
+        
+        managedObjectContext.perform {
```
seanrucker, line 59: This is used quite often, perhaps extract it to an extension on `NSManagedObjectContext`?

Sources/Data/Model/Notification.swift
```diff
@@ -0,0 +1,139 @@
...
+            "id": self.id.uuidString,
+            "campaignID": campaignID
+        ]
+    }
+}
+
+extension Notification {
+    public func markRead() {
+        guard let managedObjectContext = self.managedObjectContext else {
+            assertionFailure("Not associated with a managed object context, cannot delete.")
```
seanrucker: Same issue as before... "delete"?

Sources/UI/NotificationCenter/NotificationCenterViewController.swift
```diff
@@ -424,3 +346,152 @@ extension NotificationCenterViewController: UIViewControllerTransitioningDelegat
         return SlideRightAnimator()
     }
 }
+
+
+public class NotificationsDataSource: NSObject, UITableViewDataSource {
```
seanrucker: Instead of creating a separate class for this, why not just make your `NotificationCenterViewController` implement `UITableViewDataSource`. You didn't really gain anything from making it a separate class that is only ever used in one place. Instead you've added an additional layer that makes it slightly more difficult to reason about, adds an additional item that needs to be injected and more difficult for developers to override/customize.

Sources/Data/EventPipeline.swift
```diff
@@ -11,65 +11,28 @@ import CoreData
 import os
 
 public class EventPipeline {
```
seanrucker, line 13: This class doesn't seem very useful anymore. Its entire purpose is to provide a convenience function that requires resolving two services. Perhaps this could just be an extension to Rover itself?

Sources/UI/NotificationCenter/NotificationCenterViewController.swift
```diff
@@ -424,3 +346,152 @@ extension NotificationCenterViewController: UIViewControllerTransitioningDelegat
...
+            os_log("Problem fetching notifications list: %s", log: .ui, error.localizedDescription)
+        }
+    }
+    
+    /**
+     Returns a filtered list of notifications from dataStore.notifications. The default implementation filters out notifications that aren't notification center enabled, are deleted or have expired. This method is called automatically by the NotificationCenterTableView and the results are cached for performance.
+     
+     You can override this method if you wish to modify the rules used to filter notifications. For example if you wish to include expired notifications in the table view and instead show their expired status with a visual indicator.
+     */
+    open var notificationFilterPredicate: NSPredicate {
```
seanrucker: Looks like this can be removed as you are now using the predicate property on line 368 and this is not referenced anywhere.

Sources/UI/NotificationCenter/NotificationCenterViewController.swift
```diff
@@ -424,3 +346,152 @@ extension NotificationCenterViewController: UIViewControllerTransitioningDelegat
...
+        return fetchedResultsController.object(at: indexPath) as RoverData.Notification
+    }
+    
+    open func cellReuseIdentifier(at indexPath: IndexPath) -> String {
+        return "notification"
+    }
+    
+    // MARK: UITableViewDataSource
+    
+    open func numberOfSections(in tableView: UITableView) -> Int {
```
seanrucker: Since our implementation does not use sections, you can omit this method. The default will return 1 which is what we want. If a developer's implementation does something different they can override this themselves.

Sources/UI/NotificationCenter/NotificationCenterViewController.swift
```diff
@@ -424,3 +346,152 @@ extension NotificationCenterViewController: UIViewControllerTransitioningDelegat
...
+        guard let notificationCell = cell as? NotificationCell else {
+            return cell
+        }
+        
+        let notification = self.notificationAt(indexPath: indexPath)
+        notificationCell.configure(with: notification, imageStore: self.imageStore)
+        return notificationCell
+    }
+    
+    open func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
```
seanrucker: Similar to comment above, these three secant related methods can be omitted.

Sources/UI/NotificationCenter/NotificationCenterViewController.swift
```diff
@@ -424,3 +346,152 @@ extension NotificationCenterViewController: UIViewControllerTransitioningDelegat
...
+    open func sectionIndexTitles(for tableView: UITableView) -> [String]? {
+        return fetchedResultsController.sectionIndexTitles
+    }
+    
+    open func tableView(_ tableView: UITableView, sectionForSectionIndexTitle title: String, at index: Int) -> Int {
+        return fetchedResultsController.section(forSectionIndexTitle: title, at: index)
+    }
+}
+
+extension NotificationsDataSource : NSFetchedResultsControllerDelegate {
```
seanrucker: These methods all operate on the `subscribedTableView` property which from what I can tell is never set. If you take my suggestion of moving the `NotificationsDataSource` functionality to the `NotificationCenterViewController` itself this problem goes away.

Sources/UI/NotificationCenter/NotificationCenterViewController.swift
```diff
@@ -424,3 +346,152 @@ extension NotificationCenterViewController: UIViewControllerTransitioningDelegat
...
+    
+    open func controllerWillChangeContent(_ controller: NSFetchedResultsController<NSFetchRequestResult>) {
+        self.subscribedTableView?.beginUpdates()
+    }
+    
+    open func controller(_ controller: NSFetchedResultsController<NSFetchRequestResult>, didChange anObject: Any, at indexPath: IndexPath?, for type: NSFetchedResultsChangeType, newIndexPath: IndexPath?) {
+        switch(type) {
+        case .insert:
+            if let newIndexPath = newIndexPath {
+                let target = notificationAt(indexPath: newIndexPath)
```
seanrucker: `target` is never used. Xcode gives you a warning.

Sources/UI/NotificationCenter/NotificationCenterViewController.swift
```diff
@@ -424,3 +346,152 @@ extension NotificationCenterViewController: UIViewControllerTransitioningDelegat
...
+                subscribedTableView?.insertRows(at: [newIndexPath], with: .fade)
+            }
+        case .delete:
+            if let indexPath = indexPath {
+                subscribedTableView?.deleteRows(at: [indexPath], with: .fade)
+            }
+        case .update:
+            if let indexPath = indexPath {
+                let target = notificationAt(indexPath: indexPath)
```
seanrucker: `target` is never used. Xcode gives you a warning.

Sources/UI/NotificationCenter/NotificationCenterViewController.swift
```diff
@@ -424,3 +346,152 @@ extension NotificationCenterViewController: UIViewControllerTransitioningDelegat
...
+                subscribedTableView?.reloadRows(at: [indexPath], with: .fade)
+            }
+        case .move:
+            if let indexPath = indexPath, let newIndexPath = newIndexPath {
+                subscribedTableView?.moveRow(at: indexPath, to: newIndexPath)
+            }
+        }
+    }
+    
+    open func controller(_ controller: NSFetchedResultsController<NSFetchRequestResult>, didChange sectionInfo: NSFetchedResultsSectionInfo, atSectionIndex sectionIndex: Int, for type: NSFetchedResultsChangeType) {
```
seanrucker: We don't use sections, this can be omitted.

seanrucker (changes requested):

Sources/Data/Model/Event.swift
orospakr: done.

Sources/Data/EventPipeline.swift
orospakr: keeping it, usage sites would all have to skip DI.

Sources/Data/Model/Event.swift
```diff
@@ -35,3 +35,64 @@ public final class Event : NSManagedObject {
         self.timestamp = Date()
     }
 }
+
+extension Event {
+    /// Register a callback to be fired whenever Events are inserted.
+    ///
+    /// Note that this returns an opaque chit object that you must retain until you no longer wish the callback to be fired.
+    public static func observeNewEvents(
```
seanrucker: This is now a bit awkward for a developer to use. Perhaps this makes more sense on the EventPipeline since we've decided we're keeping it?

Sources/Data/Model/Event.swift
```diff
@@ -35,3 +35,64 @@ public final class Event : NSManagedObject {
...
+        }
+    }
+    
+    public func attemptInsert() {
+        guard let managedObjectContext = self.managedObjectContext else {
+            assertionFailure("Not associated with a managed object context, cannot insert.")
+            return
+        }
+        
+        managedObjectContext.perform {
```
seanrucker, line 46: Should this not be using your new `saveOrRollback` method defined in your extension?

Sources/UI/NotificationCenter/NotificationCenterViewController.swift
```diff
@@ -23,75 +25,37 @@ open class NotificationCenterViewController: UIViewController {
...
         self.websiteViewControllerProvider = websiteViewControllerProvider
         
         super.init(nibName: nil, bundle: nil)
         
-        notificationsObservation = notificationStore.addObserver { [weak self] _ in
-            DispatchQueue.main.async {
-                self?.cache = nil
-                self?.tableView.reloadData()
-            }
+        do {
```
seanrucker: I don't know if this makes sense here. Do you need to fetch results in the initializer? This probably makes more sense in the `viewDidLoad` method. At this point you haven't even assigned the controller as the delegate to the fetchedResultsController so the result of this fetch would not even be used.

seanrucker (changes requested):

seanrucker (approved):

PR #382, opened 2019-01-03
Feature/375 experience store on disk

### LINKED ISSUE #375: Move Experience data model concerns into Data module

orospakr:
Resolves #375.


Sources/Data/Model/Experiences/BarcodeBlock.swift
```diff
@@ -32,4 +32,50 @@ public struct BarcodeBlock: Block {
         self.keys = keys
         self.tags = tags
     }
+    
+    private enum CodingKeys: String, CodingKey {
```
seanrucker, line 36: You can omit the `= "..."` part for all cases where the case name is an exact match of the JSON string name. I.e. everything except for `typeName`.

Sources/Data/Model/Experiences/Experience.swift
```diff
@@ -56,6 +56,14 @@ extension Experience: Decodable {
         
         self.homeScreen = homeScreen
     }
+    
+    public func encode(to encoder: Encoder) throws {
```
seanrucker, line 60: This looks like an incomplete implementation of `encode(to:)`.

Sources/Data/Model/Experiences/Row.swift
```diff
@@ -71,6 +71,27 @@ extension Row: Decodable {
                 throw DecodingError.dataCorruptedError(forKey: CodingKeys.typeName, in: container, debugDescription: "Expected one of BarcodeBlock, ButtonBlock, ImageBlock, RectangleBlock, TextBlock, WebViewBlock – found \(typeName)")
             }
         }
+
```
seanrucker: I don't think you actually need BlockType to be Encodable since its only a utility type used when decoding a heterogeneous array of blocks.

Sources/Data/Model/Experiences/Screen.swift
```diff
@@ -20,10 +20,21 @@ public struct Screen: Decodable {
             self.style = style
             self.color = color
         }
+
```
seanrucker, line 23: Can you not use the synthesized implementation of `encode(to:)`?

Sources/Data/Services/ExperienceStore/ExperienceStore.swift
```diff
@@ -0,0 +1,22 @@
+//
+//  ExperienceStore.swift
+//  RoverData
+//
+//  Created by Sean Rucker on 2018-05-04.
+//  Copyright © 2018 Rover Labs Inc. All rights reserved.
+
+import Foundation
+
+public protocol ExperienceStore {
```
seanrucker: I'd suggest just making this a class instead of a protocol/implementation pair. One of the learnings I took from SDK 2.x is that this pattern makes it harder for developers to customize services. The counter would be that it makes it more testable, but I'd prioritize developer experience over our own test cases in priority.

Sources/Data/Services/ExperienceStore/ExperienceStoreService.swift
```diff
@@ -0,0 +1,69 @@
...
+//  Copyright © 2018 Rover Labs Inc. All rights reserved.
+//
+
+import Foundation
+import os
+
+class ExperienceStoreService: ExperienceStore {
+    func get(byID id: String) -> Experience? {
+        do {
+            guard let path = experienceLocalURL(id: id) else {
```
seanrucker: Small stylistic thing but this doesn't need to be inside the do/catch and reduces the indentation level if you pull it out.

Sources/Data/Services/ExperienceStore/ExperienceStoreService.swift
```diff
@@ -0,0 +1,69 @@
...
+                os_log("Not able to retrieve experience because local storage for it could not be determined.")
```
seanrucker: All `os_log` calls need to have a "namespace" (forget the proper term) so they can be identified as Rover noise in the Console app when developers are using it for their applications.

Sources/Data/Services/ExperienceStore/ExperienceStoreService.swift
```diff
@@ -0,0 +1,69 @@
...
+                return nil
+            }
+            let json = try Data.init(contentsOf: path)
```
seanrucker: `Data.init(...)` looks weird... suggest replace with `Data(...)`

Sources/Data/Services/ExperienceStore/ExperienceStoreService.swift
```diff
@@ -0,0 +1,69 @@
...
+            os_log("Unable to retrieve experience locally: %s", log: .persistence, type: .error, error.localizedDescription)
+            return nil
+        }
+    }
+    
+    func insert(experience: Experience) {
+        do {
+            let json =  try JSONEncoder.default.encode(experience)
+            guard let path = experienceLocalURL(id: experience.id) else {
+                os_log("Not able to store experience because local storage for it could not be determined.")
```
seanrucker: `os_log` "namespace" issue...

Sources/Data/Services/ExperienceStore/ExperienceStoreService.swift
```diff
@@ -0,0 +1,69 @@
...
+            }
+            let json = try Data.init(contentsOf: path)
+            return try JSONDecoder.default.decode(Experience.self, from: json)
+        } catch {
+            os_log("Unable to retrieve experience locally: %s", log: .persistence, type: .error, error.localizedDescription)
+            return nil
+        }
+    }
+    
+    func insert(experience: Experience) {
```
seanrucker: Perhaps this method should throw rather than swallow errors... this would allow the calling instance to determine how they want to handle errors. As it is, the caller has no idea if this succeeded or not. Another option would be to return a bool?

Sources/UI/Experiences/UI/ExperienceContainer.swift
```diff
@@ -97,25 +98,20 @@ open class ExperienceContainer: UIViewController {
...
-                
-                switch result {
-                case let .error(error, shouldRetry):
-                    container.present(error: error, shouldRetry: shouldRetry)
-                case let .success(experience):
-                    container.didFetchExperience(experience)
-                }
-            }
+        let experienceId: String
+        switch identifier {
```
seanrucker: An alternative syntax to this switch case for your consideration would be:

```swift
guard case .experienceID(let id) = identifier else {
    os_log("...")
    return
}
```

seanrucker (changes requested):

seanrucker (approved):

PR #395, opened 2019-01-17
Feature/384/campaigns data model

orospakr:


orospakr:
A few points about review:

* To deal with both Codable and NSManagedObject parenting the same classes, I effectively have to insert into the context in the Decodable implementation.  It's an unwanted side-effect, but one I'm pretty sure I can't avoid.
* You'll notice there's a mix of NSCoding and NSManagedObjects.  The underlying philosophy of choosing when to use which is as follows.  I use the NSCoding/Transformable for such classes that are only used as value objects (that is, embedded structures that have no identity of their own), and that themselves have no complexity: embedded lists, references to other objects, and so on.  This is to avoid cluttering up the namespace of the Managed Object Context too excessively.

orospakr:
We've retired this approach.

PR #400, opened 2019-01-22
Feature/396: Campaigns as pure DTO.

### LINKED ISSUE #396: Create alternative implementation of Campaigns persistence & API client
In lieu of persisting entire model as-is in Core Data, and using same objects/types for both local persistence concern and API client concern, instead this time:

* separate structs for DTO-side, ideally using (mostly) synthesized Codable implementation;
* As-needed local storage with separate managed object classes from DTO. And two separate classes: AutomatedCampaign and ScheduledCampaign.  Leave them blank for now.

(We will add storage of useful information to persisted Campaigns in a later ticket.  This design will flatten out trigger concern.  And individual fields in the trigger(s) can likely also be inlined, so we can have a minimum of complexity in Core Data.)

orospakr:
Resolves #396 .


Sources/Data/Model/AutomatedCampaign.swift
```diff
@@ -0,0 +1,14 @@
...
+//  RoverData
+//
+//  Created by Andrew Clunis on 2019-01-22.
+//  Copyright © 2019 Rover Labs Inc. All rights reserved.
+//
+
+import Foundation
+import CoreData
+
+public class AutomatedCampaign : Campaign {
```
seanrucker: Mark as `final` to be consistent with your other models.

Sources/Data/Model/ScheduledCampaign.swift
```diff
@@ -0,0 +1,14 @@
...
+//  RoverData
+//
+//  Created by Andrew Clunis on 2019-01-22.
+//  Copyright © 2019 Rover Labs Inc. All rights reserved.
+//
+
+import Foundation
+import CoreData
+
+public class ScheduledCampaign : Campaign {
```
seanrucker: Mark as `final` to be consistent with your other models.

Sources/Data/Model/Beacon.swift
```diff
@@ -14,12 +14,12 @@ public final class Beacon: NSManagedObject {
         return NSFetchRequest<Beacon>(entityName: "Beacon")
     }
     
-    @NSManaged public internal(set) var id: String
```
seanrucker, line 17: Instead of dropping `internal` (in fact these could even be `fileprivate` or `private`) consider defining a static function as the sole interface for `NSManagedObject` creation. You can group the set of "constructor" objects into a type (we discussed `InsertionInfo` as a naming convention).

Sources/Data/Model/Geofence.swift
```diff
@@ -14,12 +14,12 @@ public final class Geofence: NSManagedObject {
         return NSFetchRequest<Geofence>(entityName: "Geofence")
     }
     
-    @NSManaged public internal(set) var id: String
```
seanrucker, line 17: Same comments as Beacon.

Sources/Sync/SyncAssembler.swift
```diff
@@ -50,5 +51,42 @@ public class SyncAssembler: Assembler {
...
+    }
+    
+    public func containerDidAssemble(resolver: Resolver) {
+        let syncCoordinator = resolver.resolve(SyncCoordinator.self)!
+    
+        syncCoordinator.participants.append(resolver.resolve(GeofencesSyncParticipant.self)!)
+        syncCoordinator.participants.append(resolver.resolve(BeaconsSyncParticipant.self)!)
+        syncCoordinator.participants.append(resolver.resolve(CampaignSyncParticipant.self)!)
+        
+//        let assembler = SyncAssembler(options:Geofences|Notifications|Beacons|Campaigns)
```
seanrucker: Move these comments into your GitHub issue regarding `SyncAssembler` constructor args.

Sources/Sync/SyncCoordinator/DTO/Campaign.swift
```diff
@@ -0,0 +1,397 @@
...
+enum CompoundPredicateLogicalType : String, Decodable {
+    case or = "OR"
+    case and = "AND"
+    case not = "NOT"
+}
+
+struct ComparisonPredicate : Predicate, Decodable {
+    let keyPath: String
+    let modifier: ComparisonPredicateModifier
+    let op: ComparisonPredicateModifier
```
seanrucker: Consider using backticks to get around the reserved word issue. Also, is `ComparisonPredicateModifier` the right type for this?

Sources/Sync/SyncCoordinator/Participants/CampaignSyncParticipant.swift
```diff
@@ -0,0 +1,87 @@
...
+    }
+    
+    func nextRequest(cursor: String?) -> SyncRequest {
+        let orderBy: [String: Any] = [
+            "field": "UPDATED_AT",
+            "direction": "ASC"
+        ]
+        
+        var values: [String: Any] = [
+            "first": 10, // TODO: sane page size?
```
seanrucker, line 35: My gut says this can be much larger. Maybe park it for now but create a GitHub issue to determine average campaign size (when Gzipped) and come up with a number based on data.

Sources/Sync/SyncCoordinator/DTO/Campaign.swift
```diff
@@ -0,0 +1,397 @@
+//
+//  Campaign.swift
```
seanrucker: I think pulling this out into a separate file makes sense. Can we do the same with geofences and beacons (currently inlined in the `GeofenceSyncParticipant` file)? What you're really defining here is the structure of the "node" so perhaps a more descriptive name for the folder would be `Nodes` instead of `DTO`?

seanrucker (changes requested):

Sources/Sync/SyncCoordinator/Participants/CampaignSyncParticipant.swift
orospakr: https://github.com/RoverPlatform/rover-ios/issues/403

Sources/Data/Model/Beacon.swift
orospakr: Done.

Sources/Data/Model/Geofence.swift
orospakr: Done.

Sources/Sync/SyncCoordinator/DTO/Campaign.swift
orospakr: done

seanrucker (approved):

PR #408, opened 2019-02-05
Feature/398 experience sync support

orospakr:
Spending some time better understanding Protocol Oriented Programming was useful for getting my head around how to refactor things.

Here's my first pass.


Sources/Sync/SyncCoordinator/SyncCoordinatorService.swift
```diff
@@ -49,11 +54,13 @@ class SyncCoordinatorService: SyncCoordinator {
...
-        self.recursiveSync(participants: participants, requests: requests)
+        self.recursiveSync(participants: participants, variables: variables)
     }
     
-    func recursiveSync(participants: [SyncParticipant], requests: [SyncRequest], newData: Bool = false) {
-        if participants.isEmpty || requests.isEmpty {
+    func recursiveSync(participants: [SyncParticipant], variables: [String: Any], newData: Bool = false) {
+        // TODO: because I flattened out the former "requests" parameter into a single dict, checking it for emptiness no longer makes semantic sense here.
+        // however, by deleting that check have it broken an exit case for the algorithm?
+        if participants.isEmpty {
```
orospakr, line 59: What do you think about this change? Does it break any assumptions that you can remember?

seanrucker (changes requested):
The logic is super hard to reason about (my fault, not yours). I do think this can be greatly simplified if you took a fresh stab at it given the different set of constraints we're working with now (the set of participants is fixed and known by the coordinator). For example, the response could be a single defined type now instead of each participant decoding the data object into its own response.

That being said... I think we can leave this for now–with your confirmation its been tested and works as expected–and revisit it at a later time.

One thing I *would* like to address now is the public scope of the module. Right now the public API footprint is very big and I don't think it needs to be. I believe the only thing that needs to be public is a single method to initiate a sync and perhaps some configuration.

If we can tighten up the public scope of this module to that single method, then I'm fine to move on because it is essentially a block box that does its job and we can look at it again in the future if we decided it's warranted.

Sources/Sync/SyncAssembler.swift
```diff
@@ -65,24 +74,25 @@ public class SyncAssembler: Assembler {
...
-                context: resolver.resolve(NSManagedObjectContext.self, name: "backgroundContext")!,
+                syncStorage: CoreDataSyncStorage(context: resolver.resolve(NSManagedObjectContext.self, name: "backgroundContext")!),
                 userDefaults: UserDefaults.standard
             )
         }
     }
     
     public func containerDidAssemble(resolver: Resolver) {
-        let syncCoordinator = resolver.resolve(SyncCoordinator.self)!
+        let syncCoordinator = resolver.resolve(SyncCoordinator.self) as! SyncCoordinatorService
```
orospakr: What do you think of this downcast here?

I did this in order to keep the participants array scoped internal.

orospakr:
@seanrucker addressed review regarding scoping.  But I agree that more refactoring here is possible.

(PS burning late night oil because insomnia, nothing heroic lol)

seanrucker (changes requested):
Make these internal and we're good to go.

image

seanrucker (approved):

PR #412, opened 2019-02-21
Feature: Match Automated Campaigns from Events.

### LINKED ISSUE #389: Attempt to Map Predicate to NSPredicate (with LHS field type validations to avoid crashes)
[Predicates Spec](https://paper.dropbox.com/doc/Predicates--AU8ehAi3rEVkkEmaePY2ztopAg-QjeJg0q46adihvNybcQp8)

### LINKED ISSUE #409: For a given tracked event, query for automated campaigns that apply to the event and the device
The green item is the scope of this ticket.

There are several concerns that will go into selecting the automated campaigns that apply to the given observed event:

* Current device state, aka Segmentation;
* Filtering on event contents;
* Filtering by current time.

See the `CampaignAutomatedTrigger` type in GraphQL for more details, namely the `segment` and eventTrigger` fields.  Note that while `EventTriggerFilters` are stored as an array, only one of each type is permitted.

A way of evaluating predicates will be needed.  Rather than implementing an entire implementation of Predicate, consider mapping them to NSPredicate: #389.

Note that in any event DeviceSnapshot may likely need a solution for evaluating keyPaths so the dynamically bound predicates can work.

The Automated Campaign Model in Core Data is currently empty; introduce elements of the Campaign DTO (`CampaignNode`) being synced down into the locally stored one as needed.  Note that some of those fields will need to be queried on, so consider denormalizing/flattening as necessary.

orospakr:
This implements the entire spec for automated campaigns, including the Predicates.

* Stores all needed automated campaign data (filters, etc.) in Core Data
* Logic for applying all of the automated campaign filters to a given event (and surrounding context, using device info and time);
* subscribing to all emitted events and printing a log of which campaigns match;
* fairly exhaustive test coverage.

@seanrucker you may be interested to know how I solved the NSPredicate validation crash issue: in lieu of attempting to validate all of the NSPredicate rules by dead-reckoning their validation logic myself (and it appears there lots of rules),  I decided to take what I think is a more robust approach by writing an objc wrapper that handles such errors gracefully and treats them as non-matching, since Swift naturally can't catch them.

Resolves #389.  (mapping of Predicate to NSPredicate)
Resolves #409.


Sources/Sync/SyncCoordinator/SyncClient.swift
```diff
@@ -51,16 +50,6 @@ private let query = """
...
-        experiences(first: $experiencesFirst, after: $experiencesAfter, orderBy: $experiencesOrderby) {
-            nodes {
-                ...experienceFields
-            }
-            pageInfo {
-                endCursor
-                hasNextPage
-            }
-        }
     }
```
orospakr, line 53: leaving this disabled until experiences endpoint is deployed on production.

Sources/Sync/SyncCoordinator/Participants/CampaignSyncParticipant.swift
```diff
@@ -70,14 +70,73 @@ extension CampaignsSyncResponse: PagingResponse {
...
+            )
+            
+            _ = AutomatedCampaign.insert(into: context, insertionInfo: insertionInfo)
         default:
             fatalError("Some other type somehow appeared for CampaignTrigger")
         }
     }
 }
+
+extension Array {
```
seanrucker, line 112: Could these be made fileprivate to avoid polluting the Array API?

Sources/Sync/SyncCoordinator/SyncCoordinatorService.swift
```diff
@@ -10,6 +10,8 @@ import os.log
 import UIKit
 
 class SyncCoordinatorService: SyncCoordinator {
+    var observers = ObserverSet<UIBackgroundFetchResult>()
```
seanrucker, line 13: What is this used for?

seanrucker (changes requested):
- Using FP for `automatedCampaignsMatching` has the following tradeoffs:
  - Inconsistent with the rest of the codebase
  - Foreign, confusing style for the majority of iOS developers
  - Eschews our DI framework, restricting a developers ability to customize any part of it
- Suggest merging the `automatedCampaignsMatching` and `queryPredicateForCampaignQueryableFilters` functions into the `CampaignEventObserver` class and naming it `AutomationEngine`. The dependencies of the current `CampaignEventObserver` class and the `automatedCampaignsMatching` function both share the same dependencies: `DeviceSnapshot` and `NSManagedObjectContext`. 
- The extensions approach found in the current `AutomationEngine.swift` is good and a familiar concept with iOS developers. Consider marking them fileprivate so the API of the extended objects is restricted and not globally polluted.
- Move Havesine into the new AutomationEngine class as static funcs and constants.
- EventPipeline was changed from using a simple wrapper around `NotificationCenter` to using the `ObserverSet` utility I created. We had discussed in the past that `ObserverSet` doesn't add anything we can't do with a simple convenience wrapper around observing `NotificationCenter`. We had planned to remove the use of `ObserverSet`, not the other way around.
- Do we still need the Campaign entity in our data model now that we have separate AutomatedCampaign and ScheduledCampaign types?
- Can we delete TEMPORARY.swift now that it is empty?

orospakr:
I decided to put the Campaign entity in as an abstract super entity that polymorphs to either the scheduled or automated variant.  It has already come in handy for having a Campaign ID (without extra type information) on my scheduled notifications in the upcoming PR.

seanrucker (changes requested):
- Make the `automatedCampaignsMatching` function a static function on the `AutomatedCampaignsFilter` class. Perhaps even rename this to `CampaignsFilter` so it can house more static functions related to filtering campaigns in the future? Just a thought…
- Create a method on `AutomationEngine` named something similar (e.g. `func automatedCampaigns(matching event: Event)` which delegates out to `CampaignsFilter.automatedCampaignsMatching`. 
- This approach is a win for both of us. The  `automatedCampaignsMatching` function remains a pure function and factors out all the complex logic from the `AutomationEngine`. At the same time, if the developer wanted to change the way applicable campaigns are determined, they have an extension point on the engine. By making it a static func on a class it alleviates my concern around global functions being out of style with the rest of our SDK.
- Also make sure your header comments match the file names ;)

orospakr:
huh re the idea of bundling pure static funcs into a class, swiftlint had an opinion:

`Convenience Type Violation: Types used for hosting only static members should be implemented as a caseless enum to avoid instantiation. (convenience_type)`

I'll experiment with it. Interesting idea.

seanrucker (approved):

PR #414, opened 2019-03-07
Swiftlint

orospakr:


Sources/Experiences/UI/ScreenViewLayout.swift
```diff
@@ -45,7 +47,8 @@ class ScreenViewLayout: UICollectionViewLayout {
         needsPreparation = false
     }
     
-    open func prepare(frame: CGRect) {
+    // swiftlint:disable:next cyclomatic_complexity function_body_length
+    func prepare(frame: CGRect) {
```
orospakr, line 51: @seanrucker should this actually be open?

seanrucker: Probably not. I can’t think of a reason to override it. You’d essentially be creating a completely new layout if you did that.

seanrucker (approved):

Sources/Experiences/UI/ScreenViewLayout.swift
orospakr: 👍 boom

PR #424, opened 2019-03-20
Ripout campaigns functionality

orospakr:
* Add a bit of new functionality to allow for subscription to events via iOS Notification Center;
* Remove the old DI framework, which is now overkill for the much reduced scope of the core Rover SDK;
* Change some elements of the API to suit the new product scope.

Campaigns functionality is being moved to https://github.com/RoverPlatform/rover-campaigns-ios
222;


seanrucker (changes requested):
- [x] Remove all checks for Swift version. We can assume min version of 4.2 (i.e. Xcode 10)
- [x] Rename ExperienceNavigationViewController to just NavigationController (omit "View").

seanrucker (approved):

PR #425, opened 2019-04-02
2.4.0: Swift 5 regression: EXC_BAD_ACCESS signal when encoding gzip payloads.

orospakr:
Replace 1024jp's Gzip library with Markus Wanke's DataCompression.

Note this does not bump the project to Swift 5 source compilation mode, but it is now compatible with the Swift 5 compiler.

(Remove use of gzip library to decode responses; this is already handled by NSURLSession.)


orospakr:
Still going to do a bit more testing before releasing this.

seanrucker (approved):

orospakr:
Upstream (well, the old upstream)'s issue: https://github.com/1024jp/GzipSwift/issues/34

PR #435, opened 2019-04-30
Feature: Sample App and Opaque Campaign IDs

### LINKED ISSUE #433: Transform Debug App into Sample App.
* Demonstrate loading an experience directly from IDs;
* Demonstrate wiring up intent filters in your own activities for handling deep and universal links.

Tests:
1. Present an experience with a custom formed deep link (with and without a campaign ID). I.e. `myapp://rover/123` and  `myapp://rover/123?campaignID=foobar`
2. Present an experience with a universal link (with and without a campaign ID). I.e. `https://myapp.rover.io/abc123` and `https://myapp.rover.io/abc123?campaignID=foobar`

For both of the above tests, verify that the events being tracked include the `foobar` campaign ID when it is passed.

### LINKED ISSUE #434: Make Campaign IDs opaque and only use them for Event tracking.
This is to enable Rover 3.x to interoperate with other automation solutions.

* Remove campaign ID from GraphQL experiences query;
* Do not expect CampaignID in response.  Remember it locally instead;
* Include the provided CampaignID in tracked events.

Modify view controller API to optionally accept CampaignIDs, but require either Experience ID or URL.  Fetching an experience from only campaign id is no longer supported.  Eg.:

```swift
RoverViewController(id: String, campaignID: String?)
RoverViewController(url: URL, campaignID: String?)
```

orospakr:
Sample app with a simple implementation of handling deep and universal links (universal links untested since iTunes Connect shenanigans need to be done first).

Resolves #433 .
Resolves #434.


Sources/UI/NavigationController.swift
```diff
@@ -12,13 +12,16 @@ import UIKit
 open class NavigationController: UINavigationController {
     private let sessionController: SessionController
     public let experience: Experience
+    public let campaignId: String?
```
seanrucker: The established convention is to use ID (vs Id)

Sources/UI/RoverViewController.swift
```diff
@@ -51,17 +53,26 @@ open class RoverViewController: UIViewController {
         return cancelButton
     }()
     
-    public init(identifier: ExperienceIdentifier) {
+    public init(identifier: ExperienceIdentifier, campaignId: String? = nil) {
```
seanrucker: Can we make this initializer module scope? That way we can make `ExperienceIdentifier` module scope as well and reduce some of the public API. I believe this will make the RoverViewController API more straightforward.

Sources/Extensions/UIApplication.present.swift
```diff
@@ -0,0 +1,79 @@
+//
+//  UIApplication+present.swift
```
seanrucker: For consistency with the other extensions, the filename (and header comment) should just be `UIApplication` not `UIApplication.present`.

Sources/UI/RoverViewController.swift
```diff
@@ -51,17 +53,26 @@ open class RoverViewController: UIViewController {
...
         super.init(nibName: nil, bundle: nil)
         
         configureView()
         layoutActivityIndicator()
         layoutCancelButton()
     }
     
-    public convenience init(experienceId: String) {
-        self.init(identifier: .experienceID(id: experienceId))
+    public convenience init(experienceId: String, campaignId: String? = nil) {
```
seanrucker: Capital convention: experienceID and campaignID.

Sources/UI/RoverViewController.swift
```diff
@@ -51,17 +53,26 @@ open class RoverViewController: UIViewController {
...
+        self.init(identifier: .experienceID(id: experienceId), campaignId: campaignId)
+    }
+    
+    public convenience init?(experienceUrl: String, campaignId: String? = nil) {
```
seanrucker: Capital convention: experienceURL and campaignID.

seanrucker: Is there a reason the experienceURL is of type String instead of URL? You're unnecessarily making it a failable initializer and the API is actually more cumbersome to work with. In your Example app you receive a variable of type URL, convert it to a string and pass it to this initializer which then converts it back to a URL.

Sources/UI/ScreenViewController.swift
```diff
@@ -91,11 +94,14 @@ open class ScreenViewController: UICollectionViewController, UICollectionViewDat
     override open func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         
-        let userInfo: [String: Any] = [
+        let userInfoWithOptionalCampaignId: [String: Any?] = [
```
seanrucker: This approach is a bit odd to me. Why not make it of type [String: Any] and optionally add the `campaignID` key if it exists. That way you don't have to call `compactMapValues { $0 }` in multiple places.

seanrucker (changes requested):

Sources/UI/NavigationController.swift
orospakr: `experienceId` appears currently. would want to change this one.

Sources/UI/RoverViewController.swift
orospakr: 👍

Sources/Extensions/UIApplication.present.swift
orospakr: 👍

Sources/UI/RoverViewController.swift
orospakr: 👍

Sources/UI/RoverViewController.swift
orospakr: 👍

Sources/UI/ScreenViewController.swift
orospakr: 👍

Sources/UI/RoverViewController.swift
seanrucker: Initializer can be module scope but making ExperienceIdentifier module scope would require too many changes.

seanrucker (approved):

PR #438, opened 2019-05-01
Example App Tweaks

seanrucker:
Cleaned up some of the comments to be more consistent stylistically with the RoverCampaigns example app.

Also added a missing `return true` statement in the deep link handling code.


orospakr (approved):
Nice improvement! Thanks

PR #439, opened 2019-05-02
Multiple Deep Link Examples

seanrucker:
While exploring the idea of a convenience initializer on the RoverViewController to remove some of the boilerplate, I found it was not adding a ton of value and actually hiding some of the implementation details that are useful to a developer's understanding of how to customize the implementation to their own app.

Instead, I opted to provide an additional deep link example, demonstrating the various ways a developer can map custom URLs to Rover experiences.


orospakr (approved):
This seems reasonable to me.

I'll work with @samwoodall to make sure the Rover SDK Android sample app is consistent with this approach.

PR #450, opened 2019-05-09
Strongly Typed Notifications

### LINKED ISSUE #448: Strongly Typed Notifications
The SDK posts notifications on the default `NotificationCenter` for experience events ("Experience Viewed", "Block Tapped" etc.). The current implementation is not ideal because the `userInfo` property only includes encoded simple types like `String`s,` Int`s etc. The current implementation "encodes" the relevant data models (e.g. `Experience`) into these simple types. This is essentially making the decision about which properties are relevant for the developer instead of allowing them access to the entire model and deciding for themselves.

- [ ] Remove the current implementation and favour passing the actual models in the `userInfo` property instead of the encoded simple types.
- [ ] Use Apple’s new (since Swift 4.2) approach for `Notification.Name` constants. I.e. add the constants to the view controller that is posting the notification instead of the global `Notification.Name`.
- [ ] Similarly, use constants defined on the view controller for the `userInfo` keys.

seanrucker:
Closes #448


Example/Example/AppDelegate.swift
```diff
@@ -6,24 +6,26 @@
...
 class AppDelegate: UIResponder, UIApplicationDelegate {
-
     var window: UIWindow?
-
+    
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         // Pass your account token from the Rover Settings app to the Rover SDK.
         Rover.accountToken = "<YOUR_SDK_TOKEN>"
+        
+        observeRoverNotifications()
```
orospakr, line 23: I think we should follow this line with some discussion of the purpose, and the cases in your own app of where you might want to do something similar.

Sources/UI/ScreenViewController.swift
```diff
@@ -425,20 +425,63 @@ open class ScreenViewController: UICollectionViewController, UICollectionViewDat
...
+    /// user dismisses the view controller and a specified time passes (default is 15 seconds).
+    ///
+    /// This notification is useful for tracking the amount of time users spend viewing a screen. However if you want to
+    /// be notified immediately when a user views a screen you should use the `screenPresentedNotification`.
+    public static let screenViewedNotification = Notification.Name("io.rover.screenViewedNotification")
+    
+    /// The `ScreenViewController` sends this when a `UIView` representing a specific block somewhere within the view
+    /// controller's hierarchy was tapped by the user.
+    public static let blockTappedNotification = Notification.Name("io.rover.blockTappedNotification")
+}
```
orospakr, line 467: Is this that new convention for defining notification names that you were talking about?

Example/Example/AppDelegate.swift
```diff
@@ -83,13 +85,52 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
-            // Use our UIApplication.present() helper extension method to find the currently active view controller,
-            // and present RoverViewController on top.
             app.present(roverViewController, animated: true)
             return true
         }
         
         return false
     }
+    
+    func observeRoverNotifications() {
```
orospakr, line 101: maybe another quick comment down here too.

orospakr (changes requested):

Sources/UI/ScreenViewController.swift
seanrucker: Yup.

Example/Example/AppDelegate.swift
seanrucker: Done.

Example/Example/AppDelegate.swift
seanrucker: Done.

orospakr (approved):

PR #452, opened 2019-05-09
Background Session Tracking

### LINKED ISSUE #449: Improve Session Tracking Reliability
Use a background task to allow the 10 second timer to continue when the app is put in the background, giving the session time to complete.

seanrucker:
Closes #449

### DIFF
--- a/Sources/Services/SessionController/Session.swift
+++ b/Sources/Services/SessionController/Session.swift
@@ -6,7 +6,7 @@
 //  Copyright © 2018 Rover Labs Inc. All rights reserved.
 //
 
-import Foundation
+import UIKit
 
 class Session {
     let keepAliveTime: Int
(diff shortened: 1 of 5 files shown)


Sources/Services/SessionController/SessionController.swift
```diff
@@ -11,32 +11,37 @@ import UIKit
...
     }
     
     deinit {
-        if let observer = didBecomeActiveObserver {
-            NotificationCenter.default.removeObserver(observer)
-        }
-        if let observer = willResignActiveObserver {
-            NotificationCenter.default.removeObserver(observer)
-        }
+        observers.forEach(NotificationCenter.default.removeObserver)
```
orospakr, line 44: clear observers array afterwards?

Sources/Services/SessionController/Session.swift
```diff
@@ -72,21 +72,31 @@ class Session {
         case .started(let startedAt):
             // Capture endedAt now, as opposed to when the timer fires
             let endedAt = Date()
+            
+            var backgroundTaskID: UIBackgroundTaskIdentifier!
+            backgroundTaskID = UIApplication.shared.beginBackgroundTask(withName: "Keep Alive Timer") { [weak self] in
```
orospakr, line 78: I think a comment explaining what you just explained to me would be helpful: that we want a chance to expire our sessions without waiting for the app to be brought back into the foreground sometime in an arbitrarily distant future. If I have that right.

orospakr (changes requested):

Sources/Services/SessionController/SessionController.swift
seanrucker: Good catch.

Sources/Services/SessionController/Session.swift
seanrucker: You do, however I was explaining to you what the Apple docs arguably do a better job of. I.e. if you understand how Apple's background tasks work, then this code is very straightforward.

orospakr: I'm thinking more of how it interacts with our expiry algorithm.  eg.:

    // Best-effort attempt to expire our background sessions before app is killed in the background.

orospakr (approved):

PR #454, opened 2019-05-09
Fix Erroneous Public API

seanrucker:
These examples of unnecessary public API are in-compatible with the `RoverCampaigns` SDK.


orospakr (approved):
Doesn't seem like that needs to be user extensible. LGTM!

PR #456, opened 2019-05-11
Public API Audit

seanrucker:
This PR reduces the _public_ API footprint of the SDK. It trades extensibility and configurability in exchange for exposing only purposeful API. Anything the SDK exposes publicly should provide in-line comments to the developer explaining how and why they would want to use a provided class, method or variable. 

For this release we've decided on making the public API as small as possible. This provides us greater flexibility in the future to expand the public API in the future when we encounter real customer use cases. Expanding the API can be done with minor releases whereas restricting it would mean a breaking change.

The following changes to the SDK have been made in this PR:

- The entire HTTP folder consisted of public, flexible types and methods for general purpose networking. This was leftover from 2.x and after auditing its functionality it was determined that the only use of the functionality was to fetch experiences. Only ~20% of the code in this folder was every actually executed. The entire folder was removed and the relevant code moved closer to where it is used.
- The Vendor folder contained only a library for Gzipping data. This was also leftover from 2.x and wasn't actually used for anything.
- The code related to fetching and caching experiences was collapsed into a single, internal, singleton service named `ExperienceStore`. All sub-types related to the purposes of fetching and caching experiences were made private. The only _internal_ API exposed by the service is two methods: One to retrieve an experience from cache (if it exists) and another to fetch an experience over the network. 
- The code related to fetching and caching images was collapsed into a single, internal, singleton service named `ImageStore`. All sub-types related to the purposes of fetching and caching images were made private. The only _internal_ API exposed by the service is two methods: One to retrieve an image from cache (if it exists) and another to fetch an image over the network. 
- The `SessionController` was converted to a singleton and its sub-types were made private. The only _internal_ API exposed by the service is for registering and unregistering sessions.
- The `Analytics` service was returned to internal scope and is automatically enabled by the `RoverViewController`. This was done for two reasons: To _require_ it to be enabled and to reduce the public API footprint of the SDK for the reasons explained above.
- The API of the `RoverViewController` was changed slightly. Instead of accepting the experience id/url in the constructor, they are now passed in a separate method `loadExperience` method. The tradeoff for having an additional line of code when presenting an experience is that the `RoverViewController` can now be instantiated from a storyboard.
- The code responsible for presenting a loading scree while fetching an experience was refactored out into a separate view controller. The `LoadingViewController` can now be subclasses or replaced entirely making the loading screen much easier for developers to customize to match their app's look and feel.
- The `RoverViewController`, `ExperienceViewController` and `ScreenViewController` all had their public API audited. The only public/open methods now are all purposefully exposed with inline comments explaining why a developer might want to override and customize a methods default behaviour. 
- The `UITableViewCell` classes for the various block types are now internal scope. A developer can easily provide a different `UITableViewCell` implementation for a given block type (explanation of how to do this is in the `ScreenViewController` comments) but they can't extend the Rover default implementation. This greatly reduced the API footprint. If/when in the future we find a real customer use case for extending our default implementation we can revisit this.


seanrucker:
@orospakr this is quite a large PR which is why I provided such a detailed explanation of what I changed. I'm going to use my administrative super-powers to merge this in as I want to do a full release of 3.0.0-beta.1 today and set you up for final testing on Monday. I have thoroughly tested this release in both the example app as well as the Rover Campaigns example app.

orospakr:
@seanrucker hey do users still have an opportunity to modify an Experience as it passes through?

seanrucker:
Hmm... I didn’t think of that. Looks like the answer is yes, by overriding the `experienceViewController(experience:)` method, however this isn’t a very obvious way to do it. I’d say we could improve this but don’t think it would be a release showstopper. Thoughts?

orospakr:
Hey @seanrucker so @samwoodall and I have been thinking about this.  On Android, there won't be a way, obvious or otherwise, to get access to the right override point without us punching a huge hole of visibility down through the SDK's stack.  That said, I agree that locking everything down now and adding extension points later as needed gives us the greatest flexibility going forward.

So I am down with locking everything down now and deferring this to a later minor release, if needed.

PR #459, opened 2019-05-24
Carthage Support

seanrucker:
Although the Xcode scheme was indeed shared in my local environment, it was not committed to version control and therefore the SDK could not be installed via Carthage.


seanrucker:
@orospakr I'm going to go ahead and merge this since it's incredibly simple and it is holding me up.

PR #460, opened 2019-05-24
Swift 4.0 Support

seanrucker:
The SDK supported both Swift 4.2 and Swift 5.0 mode but did not support Swift 4.0 mode. Using conditional compilation we can achieve support for all three. 

In order for CocoaPods to support multiple versions we need to remove the `swift_version` entry from the podspec which will allow CocoaPods to use the same version as your project target.


orospakr (approved):
Worked well locally.  Tested session controller, barcodes, and status bar styling.  No regressions in both language modes.

PR #462, opened 2019-05-30
Allow for custom behaviour on Button Blocks

orospakr:
Requirement:

* Allow for custom integration Block Tapped handlers by customers where they are not able to use a deep link as the target.

Currently, the SDK only emits Block Tapped events if the block has an Action other than None assigned.  Indeed, we only mark those blocks that have an Action set as Clickable/Selectable to the Android & iOS UI frameworks.  Setting all the blocks as clickable (so as to allow us to pick up the click events and then dispatch our Block Tapped events), however, is no solution because it defeats the purpose of Views being set clickable, thus breaking a11y (screenreader and the like) and potentially other functionality.

In the future, we may consider adding a "Custom" action type that must be selected instead of None in order to receive the events.  In the meantime, we will instead assume that Buttons are always meant to be clickable, and only mark other Block types as clickable if they have a non-None action type set.

### DIFF
--- a/Sources/Models/BlockTapBehavior.swift
+++ b/Sources/Models/BlockTapBehavior.swift
@@ -8,7 +8,7 @@
 
 import Foundation
 
-public enum BlockTapBehavior {
+public enum BlockTapBehavior: Equatable {
     case goToScreen(screenID: String)
     case none
     case openURL(url: URL, dismiss: Bool)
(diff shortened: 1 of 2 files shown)


seanrucker (changes requested):
I would just remove that method override entirely. It's not required as the default implementation already returns true.

FYI:
> If you do not implement this method, the default return value is true.

https://developer.apple.com/documentation/uikit/uicollectionviewdelegate/1618095-collectionview

seanrucker (approved):

orospakr:
We are changing the approach here.  See PR description.

orospakr:
Replaced by #463.

PR #463, opened 2019-05-30
v3.0.0-beta5: Allow for custom behaviour on Button Blocks

### LINKED ISSUE #464: Allow for custom behaviour on Button Blocks
Requirement:

* Allow for custom integration Block Tapped handlers by customers where they are not able to use a deep link as the target.

Currently, the SDK only emits Block Tapped events if the block has an Action other than None assigned.  Indeed, we only mark those blocks that have an Action set as Clickable/Selectable to the Android & iOS UI frameworks.  Setting all the blocks as clickable (so as to allow us to pick up the click events and then dispatch our Block Tapped events), however, is no solution because it defeats the purpose of Views being set clickable, thus breaking a11y (screenreader and the like) and potentially other functionality.

So:

- [x] Mark blocks as clickable/selectable when they are either: a button, or have an action set.  This gives a nice tradeoff if making sure we don't spam all the blocks as clickable making a11y tools unusable but equally customers will be able to have clickable blocks that are only handled by their own custom logic.

In the future, we may consider adding a "Custom" action type that must be selected instead of None in order to receive the events.  In the meantime, we will instead assume that Buttons are always meant to be clickable, and only mark other Block types as clickable if they have a non-None action type set, so:

- [x] add support for a no-op Custom action type that marks the block as clickable but carries no built-in navigation behaviour;
- [x] extend the User Agent so that graphql-gateway will be able to hide that new Custom block type from SDK 2.0

New user agent will look like:

`Example/3.0.0-beta.5 CFNetwork/978.0.7 Darwin/18.6.0 iOS/12.3.1 RoverSDK/3.0.0-beta.5`

orospakr:
Resolves #464.

### DIFF
--- a/Example/Example/Info.plist
+++ b/Example/Example/Info.plist
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>3.0.0-beta.4</string>
+	<string>3.0.0-beta.5</string>
 	<key>CFBundleURLTypes</key>
 	<array>
 		<dict>
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ The recommended way to install the Rover SDK is via [Cocoapods](http://cocoapods
 Add the Rover dependency to your Podfile.
 
 ```ruby
-pod 'Rover', '~> 3.0.0-beta.4'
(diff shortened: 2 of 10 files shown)


seanrucker (approved):
👏

orospakr:
Note: unfortunately, because of the way iOS bundle short version name identifiers work, any `-beta` and similar tags will be stripped off.

PR #481, opened 2019-07-17
Feature: Polls Display Mega PR

### LINKED ISSUE #468: Simple Rendering of Text Polls (creating the basic vertical of Polls in the UI subsystem)
The goal here is to model the display of the text poll block, and all the options (aka answers to be selected between) therein.

- [x] Create the View and ViewModel needed for the Text Poll View Block;
- [x] Create the View and ViewModel needed for the Options therein.
- [x] just like barcode block type, polls blocks are always auto-height.  Have it measure intrinsic height on basis of how many Options are in the poll, making use of the text measurement stuff (even though for now we're not doing any poll-level custom typography, just use a default). Take into account the height and spacing fields from the option style.

Bind it all together!  You'll have to use stubbed data somehow to do rendering.

What is *out* of scope:

* the "result display" modality
* handling selection and submission of results
* styling
* typography
* background images

However, if you want to architect it a little bit differently (for example, if you want to experiment with a better way of breaking up the view models) or exceed some of the constraints given here, you definitely have some latitude to do that if you feel like there might prove to be useful.

### LINKED ISSUE #469: Text Polls visual presentation
For poll box itself and for options:

- [x] mode for displaying final result (animation not yet needed but be aware that transition will be added);
- [x] styling (spacing, colours, etc.);
- [x] typography;
- [x] corner radius;
- [x] background images (look in ViewBackground for inspiration, possible re-use).

### LINKED ISSUE #470: Simple Rendering of Image Polls
The goal here is to model the display of the text poll block, and all the options (aka answers to be selected between) therein.

While contained by the Rover layout system (and thus ultimately constrained by width), you can use whatever tools you like (including stock Android layout stuff, like ConstraintLayout) within if you can make it work in that context.

- [x] Create the View and ViewModel needed for the Text Poll View Block;
- [x] Create the View and ViewModel needed for the Options therein.
- [x] just like barcode block type, polls blocks are always auto-height.  Have it measure intrinsic height on basis of how many Options are in the poll (2 vs 4), making use of the text measurement stuff (even though for now we're not doing any poll-level custom typography, just use a default). Take into account the height and spacing fields from the option style.  Note that this is a little bit more complicated than it was for Text polls, since there is expansion behaviour, because the image options should always have a 1:1 aspect (square) ratio.

Bind it all together!  You'll have to use stubbed data somehow to do rendering.

What is *out* of scope:

* the "result display" modality
* handling selection and submission of results
* styling
* typography
* background images

However, if you want to architect it a little bit differently (for example, if you want to experiment with a better way of breaking up the view models) or exceed some of the constraints given here, you definitely have some latitude to do that if you feel like there might prove to be useful.

### LINKED ISSUE #471: Image Polls visual representation
For poll box itself and for options:

- [x] Caption box;
- [x] mode for displaying final result (animation not yet needed but be aware that transition will be added);
- [x] styling (spacing, colours, etc.);
- [x] corner radius;
- [x] typography;
- [x] background images (look in ViewBackground for inspiration, possible re-use).

### LINKED ISSUE #474: Implement Poll Tap and Reveal Animation
As specified on the Android ticket RoverPlatform/rover-android#391 .

### LINKED ISSUE #479: Update polls models to reflect GraphQL experience response
The GraphQL experience response has changed so the sdk models and serialization/deserialization logic in the sdk need to be updated to reflect this.
Related PR is https://github.com/RoverPlatform/rover/pull/3444

orospakr:
Sorry about this. A fair amount of non-reviewed WIP accumulated over the last few weeks. This is (largely) all of the display concerns for Polls.

Introduces:

* Polls block data types & serialization thereof;
* The custom views for both text and image polls;
* The animation and results state for both.

Resolves #468;
Resolves #470;
Resolves #471;
Resolves #474;
Resolves #469;
Resolves #479.

Left out:

* The famous result indicator.


Example/Example/ViewController.swift
```diff
@@ -7,11 +7,32 @@
...
+        decoder.dateDecodingStrategy = .formatted(DateFormatter.rfc3339)
+        
+        let experience = try! decoder.decode(Experience.self, from: experienceJson)
+        
+        
+        let rvc = RoverViewController.init()
+        rvc.loadExperience(experience: experience)
+        
+        self.present(rvc, animated: false)
+    }
```
orospakr, line 35: this is in here temporarily during the Polls project. I will be reverting it before release.

Example/Example/experience.json
```diff
@@ -0,0 +1,355 @@
...
+        },
+        "useDefaultStyle": false
+      },
+      "keys": {},
+      "tags": []
+    }
+  ],
+  "keys": {},
+  "tags": []
+}
```
orospakr: this will also be removed before release. just makes working on the test bench easier.

Sources/Extensions/DateFormatter.swift
```diff
@@ -9,7 +9,7 @@
 import Foundation
 
 extension DateFormatter {
-    static let rfc3339: DateFormatter = {
+    public static let rfc3339: DateFormatter = {
```
seanrucker, line 12: Why is this public?

Sources/Models/Image.swift
```diff
@@ -10,15 +10,13 @@ import Foundation
 
 public struct Image: Decodable {
     public var height: Int
-    public var isURLOptimizationEnabled: Bool
     public var name: String
     public var size: Int
     public var width: Int
     public var url: URL
     
-    public init(height: Int, isURLOptimizationEnabled: Bool, name: String, size: Int, width: Int, url: URL) {
```
seanrucker: Why was this removed?

Sources/Services/ImageStore.swift
```diff
@@ -176,7 +188,7 @@ extension ImageStore.Configuration {
         }
         
         let optimization: ImageStore.Optimization?
-        if image.isURLOptimizationEnabled {
+        if true {
```
seanrucker: Why?

Sources/UI/RoverViewController.swift
```diff
@@ -56,6 +56,13 @@ open class RoverViewController: UIViewController {
         loadExperience()
     }
     
+    /// Present an Experience directly into the view controller without downloading one by an identifier.
+    public func loadExperience(experience: Experience) {
+        let viewController = experienceViewController(experience: experience)
+        setChildViewController(viewController)
+        return
```
seanrucker, line 65: You need to call `self.setNeedsStatusBarAppearanceUpdate()` after you set the child view controller. Take a look at the code that is executed after a successful load (line 95). You may want to refactor it.

Sources/UI/RoverViewController.swift
```diff
@@ -56,6 +56,13 @@ open class RoverViewController: UIViewController {
         loadExperience()
     }
     
+    /// Present an Experience directly into the view controller without downloading one by an identifier.
+    public func loadExperience(experience: Experience) {
```
seanrucker: This should support passing in an optional `campaignID` for consistency with the other methods of loading/presenting experiences.

Sources/UI/Views/Extensions/UIImageView.swift
```diff
@@ -0,0 +1,65 @@
...
+//
+//  Created by Andrew Clunis on 2019-07-22.
+//  Copyright © 2019 Rover Labs Inc. All rights reserved.
+//
+
+import UIKit
+
+extension UIImageView {
+    // swiftlint:disable:next cyclomatic_complexity // This routine is fairly readable as it is, so we will hold off on refactoring it, so silence the complexity warning.
+    func configureAsBackgroundImage(background: Background?, checkStillMatches: @escaping () -> Bool = { true }) {
```
seanrucker, line 13: The `checkStillMatches` approach is a bit odd but I can live with it since it's only internal.

Sources/UI/Views/Extensions/UIImageView.swift
```diff
@@ -0,0 +1,65 @@
...
+            if case .tile = background.contentMode {
+                self.backgroundColor = UIColor(patternImage: image)
+            } else {
+                self.image = image
+            }
+            self.alpha = 1.0
+        } else {
+            let originalFrame = self.frame
+            ImageStore.shared.fetchImage(for: background, frame: frame) { [weak self] image in
+                guard let image = image, checkStillMatches(), self?.frame == originalFrame else {
```
seanrucker, line 49: What is the purpose of the `self?.frame == originalFrame`? Is this to safeguard against the device being rotate while the image is being fetched? Is so, is setting no background preferable to a "stretched" one?

Sources/UI/Views/PollQuestionView.swift
```diff
@@ -0,0 +1,47 @@
...
+//
+//  Created by Andrew Clunis on 2019-07-09.
+//  Copyright © 2019 Rover Labs Inc. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+class PollQuestionView: UIView {
+    private let backgroundView = UIImageView()
```
seanrucker: Is this actually used for anything?

Sources/UI/Views/PollQuestionView.swift
```diff
@@ -0,0 +1,47 @@
...
+//  Rover
+//
+//  Created by Andrew Clunis on 2019-07-09.
+//  Copyright © 2019 Rover Labs Inc. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+class PollQuestionView: UIView {
```
seanrucker: It seems this is a regular UIView which embeds a UITextView and pins all the edges. Perhaps it could just extend from UITextView directly?

Sources/UI/Views/TextPollCell.swift
```diff
@@ -0,0 +1,332 @@
...
+    }
+    
+    private var questionView: PollQuestionView?
+    
+    private var temporaryTapDemoTimer: Timer?
+    
+    override func configure(with block: Block) {
+        super.configure(with: block)
+     
+        containerView.translatesAutoresizingMaskIntoConstraints = false
```
seanrucker: This is redundant as `BlockCell` already sets it in the initializer.

Sources/UI/Views/TextPollCell.swift
```diff
@@ -0,0 +1,332 @@
...
+        questionView = PollQuestionView(questionText: textPollBlock.textPoll.question)
+        containerView.addSubview(questionView!)
+        questionView?.topAnchor.constraint(equalTo: containerView.topAnchor).isActive = true
+        questionView?.leadingAnchor.constraint(equalTo: containerView.leadingAnchor).isActive = true
+        questionView?.trailingAnchor.constraint(equalTo: containerView.trailingAnchor).isActive = true
+        self.optionViews = textPollBlock.textPoll.options.map { option in
+            // TODO: get initial state synchronously from the local VotingService.
+            TextPollOptionView(option: option, initialState: .waitingForAnswer)
+        }
+        for optionViewIndex in 0..<optionViews.count {
```
seanrucker: This might be a good candidate to replace with a `UIStackView`.

Sources/UI/Views/TextPollCell.swift
```diff
@@ -0,0 +1,332 @@
...
+        questionView?.removeFromSuperview()
+        self.optionViews.forEach { $0.removeFromSuperview() }
+        
+        guard let textPollBlock = block as? TextPollBlock else {
+            return
+        }
+    
+        questionView = PollQuestionView(questionText: textPollBlock.textPoll.question)
+        containerView.addSubview(questionView!)
+        questionView?.topAnchor.constraint(equalTo: containerView.topAnchor).isActive = true
```
seanrucker: When creating and activating a number of constraints at the same time, consider using the static method: `NSLayoutConstraint.activate(_:)`.

Sources/UI/Views/ImagePollCell.swift
```diff
@@ -0,0 +1,368 @@
+//
+//  ImagePollCell.swift
```
seanrucker, line 2: Same comments as `TextPollCell`. Most notably, consider using `UIStackView` to simplify layout logic.

seanrucker (changes requested):

Sources/Extensions/DateFormatter.swift
orospakr: ah, I did that to decode an experience for testing in ViewController.swift.  Couldn't set up a JSONDecoder properly otherwise.

Naturally can revert this bit after the poll project, too.

Sources/Models/Image.swift
orospakr: an early JSON I was using omitted the field causing parsing errors.  This shouldn't have persisted. I've restored it.

Sources/Services/ImageStore.swift
orospakr: fixed.

Sources/UI/RoverViewController.swift
orospakr: nice catch, thank you.

Sources/UI/Views/Extensions/UIImageView.swift
orospakr: The idea was to factor out the concern of dropping image callbacks that no longer apply to recycled views, since that arrangement doesn't fit usage outside of a collection view.

Sources/UI/Views/Extensions/UIImageView.swift
orospakr: This is at least partly to filter out stale callbacks coming back from dispatched requests from when the view was a different size.

In general, the resize event will have caused a fresh request to be dispatched, so we shouldn't be looking at a blank image.

Sources/UI/Views/PollQuestionView.swift
orospakr: nice catch. whoops.

Sources/UI/Views/PollQuestionView.swift
orospakr: Because I'm using `configureContent` to set the insets, which are done with constraints.

If I want to drop the containing UIView I would have to adopt another code path. Which, to be fair, might be as simple as using `UIView.layoutMargins` instead of constraints.  What do you think?

seanrucker: I see that but from what I can tell, there are no insets..  you're passing a hardcoded value of `.zero` which just pins the edges exactly to each other which, unless I'm missing something, makes the subview useless?

seanrucker (changes requested):

seanrucker (approved):

PR #485, opened 2019-08-14
Feature: Casting votes on Polls (along with local state persistence and final touches on animation and layout).

orospakr:


Sources/Extensions/Dictionary.swift
```diff
@@ -0,0 +1,27 @@
...
+//  Dictionary.swift
+//  Rover
+//
+//  Created by Andrew Clunis on 2019-08-01.
+//  Copyright © 2019 Rover Labs Inc. All rights reserved.
+//
+
+import Foundation
+
+extension Dictionary {
```
seanrucker: These extensions seem unnecessary. Unless I'm missing something, reduce already provides everything you need. 

```swift
let result1 = try dictionary.mapValuesWithKey { key, value in
    Int(key) != nil && value > 0
}

let result2 = dictionary.reduce(into: [:]) { result, element in
    result[element.key] = Int(element.key) != nil && element.value > 0
}

print(result1 == result2) // true
```

Sources/Services/PollsVotingService.swift
```diff
@@ -0,0 +1,387 @@
...
+private let POLLS_SERVICE_ENDPOINT = "https://polls.rover.io/v1/polls/"
+private let USER_DEFAULTS_STORAGE_KEY = "io.rover.Polls.storage"
+
+class PollsVotingService {
+    /// The shared singleton Voting Service.
+    static let shared = PollsVotingService()
+    
+    // MARK: Types
+    
+    public struct OptionStatus {
```
seanrucker: Do these two types need to be public?

Sources/Services/PollsVotingService.swift
```diff
@@ -0,0 +1,387 @@
...
+        let voteCount: Int
+    }
+    
+    public enum PollStatus {
+        case waitingForAnswer
+        case answered(resultsForOptions: [String: OptionStatus])
+    }
+
+    /// Cast a vote on the poll.  Naturally may only be done once.  Synchronous, fire-and-forget, and best-effort. Any subscribers will be instantly notified (if possible) of the update.
+    func castVote(pollId: String, givenOptionIds optionIds: [String], optionId: String) {
```
seanrucker: The lowercase "d" on `pollId` (and all others) is out of style. It's not idiomatic Swift and doesn't match the style choice in the rest of our code.

Sources/UI/Views/PollCellAnswerDelegate.swift
```diff
@@ -0,0 +1,15 @@
...
+//  PollCellAnswerDelegate.swift
+//  Rover
+//
+//  Created by Andrew Clunis on 2019-08-14.
+//  Copyright © 2019 Rover Labs Inc. All rights reserved.
+//
+
+import Foundation
+
+protocol PollCellAnswerDelegate: AnyObject {
```
seanrucker: To be more idiomatic:
• Break this up into two separate protocols
• Name each protocol after the cell name and append `Delegate`. E.g. `ImagePollCellDelegate`.
• Place each protocol definition in the same file as its corresponding cell.

Sources/UI/Views/ImagePollCell.swift
```diff
@@ -52,10 +54,15 @@ class ImagePollOptionView: UIView {
         return self.option.leftMargin
     }
     
+    public var optionId: String {
```
seanrucker: Erroneous public?

Sources/UI/Views/ImagePollCell.swift
```diff
@@ -19,6 +19,7 @@ private let RESULT_FILL_BAR_VERTICAL_SPACING = CGFloat(8)
 private let RESULT_PERCENTAGE_FONT_SIZE = CGFloat(16)
```
seanrucker, line 19: There's a lot going on in this file, it's very hard to reason about. It could be broken down into much smaller components which would help with code review. I don't think I'm suggesting a change at this point but lets discuss in-person tomorrow.

Sources/UI/Views/ImagePollCell.swift
```diff
@@ -234,18 +307,19 @@ class ImagePollCell: BlockCell {
...
     
-    private var temporaryTapDemoTimer: Timer?
-    private var temporaryTapDemoTimer1: Timer?
+    private var pollSubscription: AnyObject?
     
-    override func configure(with block: Block) {
-        super.configure(with: block)
-        self.temporaryTapDemoTimer?.invalidate()
-        self.temporaryTapDemoTimer1?.invalidate()
+    override func configure(with block: Block, for experience: Experience) {
```
seanrucker: This seems excessive to pass the experience to every block type when the only ones that need it are the two poll blocks. You could create a different configure function just for those two blocks. I also wonder if there is a simpler way to get the ID since the only use of passing the experience in here is to concatenate its ID to the blockID. This could perhaps be done in the decoding implementation.

seanrucker (changes requested):

PR #494, opened 2019-08-16
Marathon Polls UI Refactor & Polls Behaviour State Machine

orospakr:


Sources/Extensions/UserDefaults+PollsStorage.swift
```diff
@@ -0,0 +1,72 @@
...
+
+import os
+import Foundation
+
+
+private let USER_DEFAULTS_STORAGE_KEY = "io.rover.Polls.storage"
+
+extension UserDefaults {
+    
+    struct PollStorageRecord<J: Codable>: Codable {
```
seanrucker: Convention is to use T (then U, V etc.) if you don't have a descriptive name for your generic. Also, in recent versions of Swift generic constraints are preferred through the "where" syntax. 

```swift
struct PollStorageRecord<T>: Codable where T: Codable {
}
```

Sources/Extensions/UserDefaults+PollsStorage.swift
```diff
@@ -0,0 +1,72 @@
...
+        } catch {
+            os_log("Unable to update local poll storage: %s", log: .rover, type: .error, error.debugDescription)
+            return
+        }
+        os_log("Updated local state for poll %s.", log: .rover, type: .debug, id)
+        
+    }
+    
+    func retrieveStateJsonForPoll<J: Codable>(id: String) -> J? {
+        let decoder = JSONDecoder.init()
```
seanrucker: `.init()` is odd and inconsistent. Use `JSONDecoder()`.

Sources/Extensions/UserDefaults+PollsStorage.swift
```diff
@@ -0,0 +1,72 @@
...
+    struct PollStorageRecord<J: Codable>: Codable {
+        var pollID: String
+        var poll: J
+    }
+    
+    func writeStateJsonForPoll<J: Codable>(id: String, json: J) {
+        let encoder = JSONEncoder()
+        let decoder = JSONDecoder()
+        
+        let existingRecords: [PollStorageRecord<J>]
```
seanrucker: Could you repurpose your `retrieveStateJsonForPoll` method here?

```swift
let existingRecords: [PollStorageRecord<J>] = retrieveStateJsonForPoll(id: id) ?? []
```
```

File: Sources/Extensions/UserDefaults+PollsStorage.swift [hunk 4]
```diff
@@ -0,0 +1,72 @@
...
+private let USER_DEFAULTS_STORAGE_KEY = "io.rover.Polls.storage"
+
+extension UserDefaults {
+    
+    struct PollStorageRecord<J: Codable>: Codable {
+        var pollID: String
+        var poll: J
+    }
+    
+    func writeStateJsonForPoll<J: Codable>(id: String, json: J) {
```
seanrucker: Use "JSON". Also, same generic naming convention mentioned above.

```swift
func writeStateJSONForPoll<T>(id: String, json: T) where T: Codable {
}
```

Sources/Extensions/UserDefaults+PollsStorage.swift
```diff
@@ -0,0 +1,72 @@
...
+            self.set(newStateJson, forKey: USER_DEFAULTS_STORAGE_KEY)
+        } catch {
+            os_log("Unable to update local poll storage: %s", log: .rover, type: .error, error.debugDescription)
+            return
+        }
+        os_log("Updated local state for poll %s.", log: .rover, type: .debug, id)
+        
+    }
+    
+    func retrieveStateJsonForPoll<J: Codable>(id: String) -> J? {
```
seanrucker: "JSON" and generic naming convention.

Sources/Polls/URLSession+PollsRequests.swift
```diff
@@ -0,0 +1,173 @@
...
+import Foundation
+import UIKit
+import os
+
+// MARK: Network Requests
+
+private let POLLS_SERVICE_ENDPOINT = "https://polls.rover.io/v1/polls/"
+
+extension URLSession {
+    func fetchPollResults(for pollID: String, optionIds: [String], callback: @escaping (PollFetchResults) -> Void) {
```
seanrucker: ID naming convention. `optionIDs`

Sources/Polls/URLSession+PollsRequests.swift
```diff
@@ -0,0 +1,173 @@
+//
+//  URLSession+PollsRequests.swift
```
seanrucker: Since this is now the only file in the root level Polls folder, is it better suited living in the Extensions folder?

Sources/Polls/URLSession+PollsRequests.swift
```diff
@@ -0,0 +1,173 @@
...
+                let response = try decoder.decode(PollFetchResults.PollFetchResponseBody.self, from: data)
+                callback(.fetched(results: response))
+            } catch {
+                os_log("Unable to decode poll results fetch response body: %s", log: .rover, type: .error, error.debugDescription)
+            }
+        }
+        task.resume()
+    }
+
+    // DEPRECATED
```
seanrucker: If this is "deprecated" is it safe to remove since it has never been released before?

Sources/Polls/URLSession+PollsRequests.swift
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+// MARK: Voting Service REST DTOs
+
+private struct VoteCastRequest: Codable {
+    var option: String
+}
+
+enum PollFetchResults {
```
seanrucker: I would pull these two types apart so you can make one private and cleanup the type associated with the `fetched` case.

```swift
private struct PollFetchResponseBody: Codable {
    var results: [
        String: Int
    ]
}

enum PollFetchResults {
    case fetched(results: [String: Int])
    case failed
}

// Within `fetchPollResults`
let response = try decoder.decode(PollFetchResponseBody.self, from: data)
callback(.fetched(results: response.results))
```

seanrucker (changes requested):
Add your weak reference fix as well.

seanrucker (approved):

seanrucker (approved):

PR #501, opened 2019-08-19
Polls Accessibility

### LINKED ISSUE #498: Comprehensive Accessibility/TalkBack support for Poll blocks
Should probably indicate that the block is a poll, what the options are to vote on, and that it is tappable/actionable.  When voted on, should say that you have voted on it and what the percentages are.

seanrucker:
Closes #498


seanrucker:
@orospakr would be great if you could test this from a (blind) user's perspective in addition to code review.

orospakr:
Hey @seanrucker , I'm not sure if this is appropriate for Apple's Accessibility API, but some extra hint (maybe on poll title?) that established that the following things are part of a "poll", and that it has N options, might be useful.  Right now while browsing through the polls with VoiceOver, little is done to establish the contextual link between the options.  This is done visually for users who can see the display.

seanrucker:
Good call. Addressed.

orospakr (approved):
👏

PR #528, opened 2020-02-12
Hotfix v3.5.3: Ergonomics for handling Poll Answered events

orospakr:
In order to (easily) get:

* block name
* any of the poll fields

in a Poll Answered receiver without some pattern-matching/casts, there were two (obviously) missing protocol conformances that I've added in.

And in the same stroke, also added Poll Answered to the example notification receivers in the Example App's AppDelegate.


seanrucker (approved):

orospakr:
note to self: squash before running release script

Issue #529, opened 2020-03-30
Text Overflow Issue


seanrucker:
In some cases iOS incorrectly measures the size of the text field resulting in the last line being cut off.

orospakr:
Shipped in 3.6.0.

PR #530, opened 2020-04-01
Sizing and truncation issues with Text (both Text blocks and Polls)

### LINKED ISSUE #523: Poll Getting Cut Off
On the first screen of the experience with ID 5de84226b894fd000ad52be0 there is a poll at the bottom that is getting cut off on iOS. Android seems to be fine.

Screen Shot 2020-01-07 at 10 33 14 AM

(Image has been redacted).

orospakr:
Changes:

* Fixed a bug where text for plain text items (namely for polls) was always being parsed as HTML;
* Switch from using NSAttributedText's `boundingRect()` method to using a full NSTextContainer and NSLayoutManager to measuring the height of text content. This is necessary because `boundingRect()` is only a basic tool and is not fully aware of larger text flow and layout concerns, such as paragraphs.

Resolves #523, #529.


samwoodall (approved):
👍

PR #531, opened 2020-04-02
A11y fixes for Text and Actions

### LINKED ISSUE #524: Accessibility Improvements
1. Text elements are read as disabled text fields (dimmed means disabled) but should be read as plain text labels.
2. Rover blocks with tappable actions (buttons or tappable images) need to have a trait set that indicates they are tappable.

See video in Dropbox > Engineering > issues > rover-ios > 524

orospakr:
* TextBlockCell now configures its UITextView with `isEditable = false` in addition to `textView.isUserInteractionEnabled = false`. The UITextView that believed it was editable, and responded as such to accessibility requests, causing Voice Over to exclaim "Text field; dimmed" after each text block.
* The `.link` UIAccessibilityTraits bit is now set to on for blocks that have any action applied. This was done instead of `.button` because `.link` is meant for navigation tasks, and `.button` is more meant for behavioural actions (send a message, buy something, otherwise change state), which our actions do not do.

Resolves #524

### DIFF
--- a/Sources/UI/Views/BlockCell.swift
+++ b/Sources/UI/Views/BlockCell.swift
@@ -47,6 +47,7 @@ class BlockCell: UICollectionViewCell {
         configureBorder()
         configureOpacity()
         configureContent()
+        configureA11yTraits()
     }
     
     func configureBackgroundColor() {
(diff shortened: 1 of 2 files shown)


samwoodall (approved):
👍

PR #533, opened 2020-04-03
a11y actions on all block types, more a11y metadata for other block types

orospakr:
* Barcodes identify as barcodes;
* All Rover blocks are turned into a11y-enabled views;
* Some Rover blocks that lack a11y content are excluded unless they have Tap Behaviours.


samwoodall (approved):

orospakr:
Contributes to #524.

PR #534, opened 2020-06-03
Feature/conversion tracking

chrisrecalis:
Adds new conversion tracking info on blocks & screens.

Blocks & Screens now have an additional optional `conversion` property attached to them.  A `Conversion` consist of the following:
```json
{
  "tag": "String",
  "expires": {
     "unit": "ENUM",
     "value": "Int"
  }
}
```
The new information added is to be used in conversion tracking by adding additional information to the user interacting with blocks & screens for some set duration of time. This info is used in the `rover-campaigns-ios` SDK to automatically tag a user under `userInfo`. Clients can then build segments around these tags as sort of an ad-hoc behavioural segment.

### DIFF
--- a/Sources/Models/Conversion.swift
+++ b/Sources/Models/Conversion.swift
@@ -6,6 +6,16 @@
 //  Copyright © 2020 Rover Labs Inc. All rights reserved.
 //
 
+public struct Conversion: Decodable {
+    public var tag: String
+    public var expires: Duration
+    
+    
+    public init(tag: String, expires: Duration) {
+        self.tag = tag
+        self.expires = expires
+    }
+}
 
 public struct Duration: Decodable {
     public enum Unit: String, Decodable {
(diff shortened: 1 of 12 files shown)


Sources/Models/Conversion.swift
```diff
@@ -0,0 +1,53 @@
...
+            base = 1
+        case .minutes:
+            base = 60
+        case .hours:
+            base =  60 * 60
+        case .days:
+            base = 24 * 60 * 60
+        }
+        return TimeInterval(self.value * base)
+    }
```
chrisrecalis: Should this belong in `rover-campaigns-ios` as an extension rather on the model itself?

chrisrecalis: If extension is possible

orospakr: @chrisrecalis if you're only using it in rover-campaigns-ios, then yeah, you can totally do it with an extension there instead. Having it here means anything using a Conversion value from rover-ios can calculate the TimeInterval value. It's up to you!

orospakr (approved):

Sources/Models/Conversion.swift
chrisrecalis: @orospakr what do you think is the best solution? I attempted to add an extension in `rover-campaigns-ios` but wasn't able to

Sources/Models/Conversion.swift
```diff
@@ -0,0 +1,53 @@
...
+}
+
+
+public struct Conversion: Decodable {
```
orospakr: actually it's better to put the type the file is named after at the top.

PR #535, opened 2020-07-08
Add missing foundation import to model conversions

chrisrecalis:


Sources/Models/Conversion.swift
```diff
@@ -6,6 +6,8 @@
 //  Copyright © 2020 Rover Labs Inc. All rights reserved.
 //
 
+import Foundation
+
```
chrisrecalis, line 10: Does this need to be surrounded by
```swift 
#if !COCOAPODS
#endif
```

PR #536, opened 2020-07-17
Crash when submitting Poll after Editing

orospakr:
If poll is edited in the Experiences web tool after it has been displayed once with the SDK, it could crash when you attempt to submit an answer. This would occur because the locally seeded (cached) results for that poll would no longer match the poll options.

The behaviour now is to drop that vote attempt, and reset the poll back to its initial state to prompt a fresh re-load of the poll results.

### Repro Instructions:

1. Open poll in an app embedding the SDK;
2. Change one of the poll options in Experience editor;
3. Restart the app, and re-open the Experience;
4. Try to submit vote.


orospakr:
Dropping this branch to use a properly named git-flow one instead.

PR #537, opened 2020-07-17
Crash when submitting Poll after Editing 

orospakr:
If poll is edited in the Experiences web tool after it has been displayed once with the SDK, it could crash when you attempt to submit an answer. This would occur because the locally seeded (cached) results for that poll would no longer match the poll options.

The behaviour now is to drop that vote attempt, and reset the poll back to its initial state to prompt a fresh re-load of the poll results.

### Repro Instructions:

1. Open poll in an app embedding the SDK;
2. Change one of the poll options in Experience editor;
3. Restart the app, and re-open the Experience;
4. Try to submit vote.

### DIFF
--- a/Sources/UI/Views/Polls/PollCell.swift
+++ b/Sources/UI/Views/Polls/PollCell.swift
@@ -277,6 +277,8 @@ class PollCell: BlockCell {
             return true
         case (.submittingAnswer, .refreshingResults):
             return true
+        case (.submittingAnswer, .initialState):
+            return true
         case (.refreshingResults, .refreshingResults):
             return true
         case (.refreshingResults, .initialState):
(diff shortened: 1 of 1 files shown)


seanrucker (changes requested):
It seems like the first time I try to vote after a poll has changed it does not take.

Using your repro steps:

1. Open poll in an app embedding the SDK;
1. Change one of the poll options in Experience editor;
1. Restart the app, and re-open the Experience;
1. Try to submit vote. *Notice the first tap does not register a vote.*
1. Try to submit vote a second time. *Notice the second time it works correctly.*

orospakr:
Hi @seanrucker, indeed, that's exactly the expected behaviour.

> The behaviour now is to drop that vote attempt, and reset the poll back to its initial state to prompt a fresh re-load of the poll results.

As I was mentioning at our team meeting this is the "not-100%-optimal-but-safe" solution I propose in lieu of a larger (and riskier) change to the state machine.

What do you think?

seanrucker:
Let me talk to Phil and see if this is acceptable.

orospakr:
Thank you @seanrucker 🙏

seanrucker (approved):
Ok it's not ideal but I've discussed the known issue with Phil. He can communicate the double vote workaround to clients. No sense holing up this PR, it fixes the crash which is an improvement.

:shipit:

PR #543, opened 2020-10-29
Image Accessibility

orospakr:


seanrucker (approved):
Works perfectly. I discovered Xcode's Accessibility Inspector which made testing this quite easy.

PR #544, opened 2020-10-29
Prevent Problems with UIScreen.main.scale values of 0

orospakr:
Prevent a UIScreen.main.scale value of 0 from causing problems by returning a sane default (for most devices) of 3.0. This has been observed to occur in certain customer integrations.

We only use the value for specifying the density of the image we ask for Imgix to return, so as long as the value is not 0, then the correct layout for the image occurs (with the only potential issue of the resolution not perfectly matching the screen, which is almost certainly a non-issue).

### DIFF
--- a/Sources/Extensions/UIScreen+safeScale.swift
+++ b/Sources/Extensions/UIScreen+safeScale.swift
@@ -1,0 +1,17 @@
+//
+//  UIScreen+safeScale.swift
+//  Rover
+//
+//  Created by Andrew Clunis on 2020-10-29.
+//  Copyright © 2020 Rover Labs Inc. All rights reserved.
+//
+
+import UIKit
+
+extension UIScreen {
+    var safeScale: CGFloat {
+        let value = self.scale
+        // in some cases, we have seen it occur that scale has returned 0. In that event, return a sane value.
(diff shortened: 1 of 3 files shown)


Sources/Extensions/UIScreen+safeScale.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Copyright © 2020 Rover Labs Inc. All rights reserved.
+//
+
+import UIKit
+
+extension UIScreen {
+    var safeScale: CGFloat {
+        let value = self.scale
+        // in some cases, we have seen it occur that scale has returned 0. In that event, return a sane value.
+        return value == 0 ? 3.0 : value
```
seanrucker, line 15: Have you tested this to see the effect? What happens on a device with a 2x screen scale?

seanrucker (approved):

Sources/Extensions/UIScreen+safeScale.swift
orospakr: I have done some further testing, have a look at PR description.

seanrucker: Awesome, thanks!

PR #547, opened 2021-10-29
Official SwiftPM Support (and drop Carthage support)

orospakr:
Brings SwiftPM support over the line.

* While SwiftPM support was cheekily added before, it has an issue where the RoverSDK version number for the User-Agent header would not be detected properly, and would yield `unknown`.
* Remove the top-level xcodeproj, and use SwiftPM directly to depend on the SDK itself in the parent directory
* Create a `Meta.swift` file to contain the SDK version number in lieu of the xcodeproj/plist, aligning it with the rest of our SDKs that do the same.


seanrucker (changes requested):
- [x] In the README the first line under "Swift PM" reads: "The recommended way to install the Rover SDK is via Cocoapods."
- [x] The suggested GitHub URL in the README produces an error when trying to add the package dependency. 
image

- [ ] The documentation following the link from the README is out of sync with the installation methods on the README. 
image

- [x] IIRC the Rover Campaigns SDK automatically records the SDK version of the Rover Experiences SDK. If I'm correct about this, will it work with this new `Meta.SDKVersion` approach?

orospakr:
@seanrucker hey the Rover Campaigns SDK does not use the version of the Experiences SDK. It only puts its own version version number into both the device context and in the User Agent.

orospakr:
@seanrucker I've got an update ready for docs site, but not gonna push it until this is merged and released.

seanrucker (approved):

PR #550, opened 2022-11-09
Allow device name to be set when testing push

Moux:
* Add device name property to context, and allow it to be set by the user when testing.
* Rename SettingsViewController to RoverViewController.

### DIFF
--- a/Sources/Data/Context/DeviceNameManager.swift
+++ b/Sources/Data/Context/DeviceNameManager.swift
@@ -3,6 +3,7 @@
 //  
 //
 //  Created by Biso Shrestha on 2022-11-08.
+//  Copyright © 2022 Rover Labs Inc. All rights reserved.
 //
 
 public protocol DeviceNameManager {
--- a/Sources/Data/Context/DeviceNameManager.swift
+++ b/Sources/Data/Context/DeviceNameManager.swift
@@ -1,0 +1,10 @@
+//
+//  DeviceNameManager.swift
+//  
+//
+//  Created by Biso Shrestha on 2022-11-08.
+//
+
+public protocol DeviceNameManager {
+    func setDeviceName(_ deviceName: String)
+}
--- a/Sources/Debug/RoverSettingsViewController.swift
+++ b/Sources/Debug/RoverSettingsViewController.swift
@@ -10,9 +10,10 @@ import UIKit
 
 #if !COCOAPODS
 import RoverFoundation
+import RoverData
 #endif
 
-open class SettingsViewController: UIViewController {
+open class RoverSettingsViewController: UIViewController {
     public let isTestDevice = PersistedValue<Bool>(storageKey: "io.rover.RoverDebug.isTestDevice")
     
     public private(set) var navigationBar: UINavigationBar?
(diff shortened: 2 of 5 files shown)


orospakr:
@Moux you mean "RoverSettingsViewController" ;)

Sources/Data/Context/DeviceNameManager.swift
```diff
@@ -0,0 +1,10 @@
+//
+//  DeviceNameManager.swift
+//  
+//
+//  Created by Biso Shrestha on 2022-11-08.
```
orospakr, line 5: oh ya don't forget copyright line

> Copyright © 2022 Rover Labs Inc. All rights reserved.

see other files

Sources/Debug/RoverSettingsViewController.swift
```diff
@@ -308,8 +310,29 @@ extension SettingsViewController: UITableViewDelegate {
...
+    
+    public func textFieldDidEndEditing(_ textField: UITextField) {
+        RoverFoundation.shared!.resolve(DeviceNameManager.self)?.setDeviceName(textField.text ?? UIDevice.current.name)
+    }
+    
+    public func textFieldShouldReturn(_ textField: UITextField) -> Bool {
+        textField.resignFirstResponder()
+        return true
+    }
+}
```
orospakr, line 338: haha the funny thing is if I were doing this I just woulda been like "F it" and used SwiftUI 😂 

but keeping with uikit is just fine, since technically target SDK till lower than 13 for now.

Sources/Data/Context/DeviceNameManager.swift
```diff
@@ -0,0 +1,10 @@
...
+//
+
+public protocol DeviceNameManager {
+    func setDeviceName(_ deviceName: String)
+}
```
orospakr, line 11: actually I wonder that in lieu of having this separate protocol, you could just add it to StaticContextProvider. and put the persistedvalue itself on ContextManager like the others.

Or just duplicate it into RoverSettingsViewController like was done with testdevice.

orospakr (changes requested):

Sources/Data/Context/DeviceNameManager.swift
Moux: Are we going to allow this value to be set in release apps/builds?  I wasn't sure or I would have put it into the existing DebugContextProvider.

Sources/Debug/RoverSettingsViewController.swift
Moux: Yeah, I don't see the benefit of forcing developers to add a SwiftUI dependency here.

Sources/Data/Context/DeviceNameManager.swift
Moux: Will fix.

Sources/Data/Context/DeviceNameManager.swift
orospakr: ya works in prod builds.

orospakr (approved):

